Issue #52, opened 2026-07-14
kernel: every lazy kldload path is dead — the .ko tree is deleted at image assembly


pkgdemon:
## The general bug behind the `wlan_xauth` failure

`ci/assemble-image.sh:168` deletes the entire kernel module tree:

```sh
rm -f "$ROOTFS"/boot/kernel/*.ko
```

This is intentional and correct for the kext model — but it means **every lazy `kldload` path in the kernel is dead on NextBSD.** Any subsystem that autoloads a module on first use will fail silently and non-obviously, because the recovery mechanism upstream relies on has been removed wholesale. Upstream never notices, because on stock FreeBSD the `.ko` is always there.

`wlan_xauth` is simply the first instance anyone tripped over. It produced a hard `EINVAL` on every WPA association with no useful diagnostic.

## Known call sites in net80211 alone

| Site | Autoloads |
|---|---|
| `ieee80211_proto.c:476` | `wlan_xauth` (802.1X/WPA authenticator) — **confirmed broken** |
| `ieee80211_crypto.c:370` | `wlan_ccmp` / `wlan_tkip` / `wlan_wep` / `wlan_gcmp` — currently masked, because all four happen to be in GENERIC |
| `ieee80211_ratectl.c:125` | `wlan_amrr` / `wlan_rssadapt` — `amrr` is in GENERIC, `rssadapt` is not |
| `ieee80211_scan.c:191` | scan modules |

Note how fragile the "masked" ones are: they work today only by the accident that GENERIC happens to compile them in. Any upstream change that moves one of them out of GENERIC breaks us silently.

## Ask

Audit the whole kernel for `kldload` / `*_load_module()` call sites that assume on-demand module loading, and for each one either:

- compile the module into `config/NEXTBSD`, or
- ship it as a `.kext`, or
- document explicitly why the path is unreachable on NextBSD.

Consider a CI check that greps for new autoload call sites, so this class doesn't silently regrow.

Found while re-deriving the WLAN plan against source: https://pkgdemon.github.io/nextbsd-wlan-plan.html

PR #61, opened 2026-08-01
kernel: bake in FUSEFS and the 64-bit Linuxulator (linux64 + linprocfs/linsysfs)

### LINKED ISSUE #60: kernel: bake in FUSEFS and ship the Linux ABI modules (linux64/linux/linux_common/linprocfs/linsysfs) to run AppImages
## Goal

Run AppImages on NextBSD **without** populating `/compat/linux`. AppImages carry their own payload libraries, and the modern runtime is statically linked, so the intent is to provide only the kernel-side pieces: FUSE, and the Linux ABI.

Two independent halves, with very different costs.

---

## 1. FUSEFS — compile in, one line

`fusefs` is **base, not a port** — `sys/fs/fuse/` is in the kernel tree. (Ports have `fusefs-libs`/`fusefs-libs3` and the `sysutils/fusefs-*` filesystems; those are userspace and unrelated.)

It is statically configurable, declared exactly like options `NEXTBSD` already bakes in — `sys/conf/options:269`:

```
NULLFS		opt_dontuse.h
TARFS		opt_dontuse.h
UNIONFS		opt_dontuse.h
FUSEFS		opt_dontuse.h      <- same shape
```

and `sys/conf/NOTES:1119` lists `options FUSEFS` beside `CD9660`/`MSDOSFS`. So, in `config/NEXTBSD`, next to the existing TARFS/UNIONFS/NULLFS block:

```
options 	FUSEFS
```

Two things worth noting:

- **Compiling it in sidesteps #52 entirely** — there is no `.ko` for `assemble-image.sh` to delete.
- **Nothing needs installing for squashfs.** The current AppImage `type2-runtime` statically links squashfuse; its README states *"Since the runtime is linked statically, libfuse2 is no longer required on the target system."* The kernel fuse driver plus `/dev/fuse` is the entire host-side requirement. `mount_fusefs` is base and does not appear in `nextbsd-freebsd-compat/scripts/superseded`.

---

## 2. Linux ABI — modules only; it cannot be compiled in

`sys/conf/options.amd64:18-19`:

```
#COMPAT_LINUX		opt_dontuse.h
COMPAT_LINUX32		opt_dontuse.h
```

`COMPAT_LINUX` is commented out, and `sys/conf/files.amd64` contains **no** linux entries at all. The 64-bit Linux ABI has no static build path on amd64 — upstream ships it only as `sys/modules/linux64`. So unlike FUSEFS, this cannot be a config option.

Modules required (all confirmed present in releng/15.1):

| Module | Purpose |
|---|---|
| `linux64` | amd64 Linux ABI |
| `linux` | i386 Linux ABI (32-bit, as requested) |
| `linux_common` | shared ABI code, `MODULE_DEPEND` of both |
| `linprocfs` | `/proc` — **not optional**, see below |
| `linsysfs` | `/sys` |

**32-bit specifics:** `sys/modules/linux/Makefile` self-defines `-DCOMPAT_FREEBSD32 -DCOMPAT_LINUX32` on amd64, so no kernel config change is needed for it. Its kernel-side prerequisite `options COMPAT_FREEBSD32` is already in GENERIC (`sys/amd64/conf/GENERIC:63`), which `NEXTBSD` includes.

**Build change** — `.github/workflows/build.yml:204` currently passes `NO_MODULES=yes`, so the kernel build emits zero `.ko` files:

```yaml
            KERNCONF=NEXTBSD \
            MODULES_OVERRIDE="linux linux64 linux_common linprocfs linsysfs" \
            buildkernel -j"$(nproc)"
```

`MODULES_OVERRIDE` builds only the listed modules rather than GENERIC's full set, keeping build time close to today's.

---

## 3. Blocked on #52 — building them is not sufficient

`ci/assemble-image.sh:168`:

```sh
rm -f "$ROOTFS"/boot/kernel/*.ko
```

Every `.ko` is deleted at image assembly. #52 offers three remedies — compile into `config/NEXTBSD`, ship as a `.kext`, or document the path as unreachable. **The first is unavailable here** (no static build path), and the third is not an option if we want this to work. So this needs a decision:

- **Ship them as kexts**, like the drm/nvidia ones. Complication: `linux64` has no hardware to match, so there is no IOKit personality to trigger it — it needs an unconditional/boot-load path in kextd rather than match-based loading.
- **Or carve an explicit exception** into the `rm` in `assemble-image.sh`.

This is the main design question in the ticket; everything else is mechanical.

---

## 4. Loading and mounting

NextBSD has no `rc.d`, so nothing performs `rc.d/linux`'s job. Beyond loading the modules, `linprocfs` and `linsysfs` must be **mounted**. Options: a `loader.conf.d` fragment (cuts against #180's shrink-loader.conf goal) or a LaunchDaemon that loads and mounts at boot. The LaunchDaemon is more in keeping with the launchd model.

---

## 5. Scope decision: no `/compat/linux` — recorded with its limits

Deliberate: no `emulators/linux_base-*`, kernel-side only. Recording the known consequences so they aren't rediscovered as bugs:

- **`/proc` is still required.** The AppImage runtime locates itself via `/proc/self/exe` (`AppImage/type2-runtime`, `src/runtime/runtime.c:1483`). That needs `linprocfs` mounted — conventionally at `/compat/linux/proc` — even with no Linux base installed. Hence `linprocfs` in the module list above.
- **Dynamically-linked payloads will likely fail.** AppImages bundle their *application* libraries but generally not glibc or the ELF interpreter — the design assumes the host provides them. Linuxulator resolves paths through `/compat/linux` first and then the native root, so a payload asking for `/lib64/ld-linux-x86-64.so.2` will find nothing. The statically-linked musl runtime itself should load fine; the app inside it is the question. AppImages built in "bundle everything" mode would work.
- **FUSE-under-Linuxulator is unverified.** The runtime opens `/dev/fuse` and issues Linux `mount(2)` with fstype `fuse`; whether Linuxulator translates that onto FreeBSD's fusefs is untested. If it does not, `--appimage-extract-and-run` is the fallback and bypasses FUSE completely — in which case the Linux ABI is doing all the work and FUSEFS is a bonus.

---

## Acceptance

1. `options FUSEFS` in `config/NEXTBSD`; `/dev/fuse` present on a booted image.
2. `linux`, `linux64`, `linux_common`, `linprocfs`, `linsysfs` built **and surviving image assembly** (#52).
3. A statically-linked Linux x86-64 binary executes.
4. An AppImage runs — or, failing the FUSE self-mount, `--appimage-extract-and-run` does.

Refs: #52 (`.ko` tree deleted at assembly), #180 (shrinking `loader.conf`).

pkgdemon:
Closes #60 (the bake-in half).

NextBSD ships no `/boot/kernel` `.ko` tree (`assemble-image.sh:168`) and no kld tooling — `kldload`/`kldstat`/`kldunload` are superseded by `kext*` (`nextbsd-freebsd-compat/scripts/superseded:13-16`). Anything upstream ships only as a module is therefore unreachable here. FUSE and the Linux ABI are both in that category, and together they're what blocks running AppImages.

**Additive only — no upstream file is patched.** Two new `src-overlay/conf/` fragments appended by the workflow, exactly as `files.linuxkpi_video` and `files.vfs_pivot` already are.

## FUSEFS — no overlay needed

`fs/fuse/*.c` is gated `optional fusefs` in `sys/conf/files`, so the stock option compiles it in. One line in `config/NEXTBSD`. Compiling in rather than shipping `fusefs.ko` also sidesteps #52: no `.ko` for `assemble-image.sh` to delete.

## The 64-bit Linuxulator — needs the overlay

There has never been a static path for it: `#COMPAT_LINUX` is commented out in `sys/conf/options.amd64` in 11.0, 12.0, 13.0, 14.0 and 15.1, and `files.amd64` carries no linux entries. The **32-bit** ABI had one until 14.0, removed by `e013e3693` — *"Since we have 32 and 64 bit Linux emulators, we can't build both emulators together into the kernel."* We build only the 64-bit ABI, so that collision doesn't apply.

`files.compat_linux` transcribes `sys/modules/linux64/Makefile` and `sys/modules/linux_common/Makefile` into config(8) form, following the shape releng/13.0's `files.amd64` used for the static linux32 path — 38 source entries plus five generated artefacts:

```
linux_genassym.o -> linux_assym.h -> linux_locore.o
                                  \
                     linux_vdso_gtod.o -> linux_vdso.so.o -> linux_vdso_inc.S
```

`linux_vdso_inc.S` `.incbin`s `"linux_vdso.so.o"` by that exact name, so the vDSO target keeps the `.so.o` suffix.

`linprocfs`/`linsysfs` are included because the AppImage runtime resolves itself via `/proc/self/exe`, so `/proc` must exist even with no Linux base. Both are plain filesystems — no generated files, no pseudofs dependency.

## Why static registration works

The concern with compiling in a module-only subsystem is that it registers nothing. Not the case here — `amd64/linux/linux_sysvec.c`:

```c
DECLARE_MODULE_TIED(linux64elf, linux64_elf_mod, SI_SUB_EXEC, SI_ORDER_ANY);
MODULE_DEPEND(linux64elf, linux_common, 1, 1, 1);
```

with `elf64_insert_brand_entry()` in the `MOD_LOAD` case. `DECLARE_MODULE` fires `MOD_LOAD` at its SYSINIT point for statically linked code exactly as it does on `kldload`, and `linux_common` is compiled in alongside so the dependency resolves.

## Not verified locally — CI is the test loop

I could not build this before opening the PR. The source entries are a mechanical transcription and should be fine; **the five generated-artefact rules are the risk** — translating `bsd.kmod.mk` recipes into `dependency`/`compile-with` form, particularly the vDSO's custom `ld` invocation with its own linker script. Expect one or two rounds of build fixes. Opening as draft for that reason.

## Out of scope

- **Mounting** `linprocfs`/`linsysfs` — belongs in `nextbsd-overlays` (`/private/etc/fstab`), not here.
- **`/compat/linux`** — deliberately not populated. Consequence, recorded in #60: statically-linked Linux binaries will run; dynamically-linked payloads looking for `/lib64/ld-linux-x86-64.so.2` will not.
- **Whether Linuxulator translates Linux `mount(2)` with fstype `fuse`** onto FreeBSD's fusefs is untested. If it doesn't, `--appimage-extract-and-run` bypasses FUSE entirely.

## Revert path

When the kext-conversion project lands (#52), this comes out cleanly: drop four `options` lines from `config/NEXTBSD`, delete the two overlay fragments, drop the `build.yml` step. Nothing else in the tree changed.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -191,6 +191,31 @@ jobs:
           test -f /usr/src/sys/kern/vfs_pivot.c
           echo "==> vfs.pivot wired: $(grep -c 'kern/vfs_pivot.c' /usr/src/sys/conf/files) entry; source present"
 
+      # Static Linuxulator: the 64-bit Linux ABI + linprocfs/linsysfs (#60).
+      # Upstream ships these only as sys/modules/*, and there has never been a
+      # static path for the 64-bit ABI (COMPAT_LINUX is commented out in
+      # sys/conf/options.amd64; files.amd64 has no linux entries). NextBSD ships
+      # no /boot/kernel .ko tree (assemble-image.sh:168) and no kld tooling
+      # (kldload/kldstat/kldunload are superseded by kext*), so a module cannot
+      # be loaded here at all -- compiling in is the only route. Sources are
+      # stock base sources already in the clone; we append the declarations and
+      # the files fragment. KERNEL build only, same rule as linuxkpi_video.
+      - name: Wire in the static Linuxulator (linux64 + linprocfs/linsysfs, #60)
+        run: |
+          set -eu
+          cat "$GITHUB_WORKSPACE/src-overlay/conf/options.compat_linux" >> /usr/src/sys/conf/options
+          cat "$GITHUB_WORKSPACE/src-overlay/conf/files.compat_linux" >> /usr/src/sys/conf/files.amd64
+          # COMPAT_LINUX must not already be live, or config(8) sees a duplicate.
+          grep -qE '^#COMPAT_LINUX' /usr/src/sys/conf/options.amd64
+          for s in amd64/linux/linux_sysvec.c amd64/linux/linux_genassym.c \
+                   amd64/linux/linux_locore.asm amd64/linux/linux_vdso_gtod.c \
+                   amd64/linux/linux_vdso.lds.s compat/linux/linux_vdso_inc.S \
+                   compat/linux/linux_elf64.c compat/linprocfs/linprocfs.c \
+                   compat/linsysfs/linsysfs.c; do
+            test -f "/usr/src/sys/$s" || { echo "MISSING /usr/src/sys/$s" >&2; exit 1; }
+          done
+          echo "==> Linuxulator wired: $(grep -c 'optional compat_linux' /usr/src/sys/conf/files.amd64) entries; sources present"
+
       - name: Build kernel (no modules)
         run: |
           cd /usr/src
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -205,8 +205,14 @@ jobs:
           set -eu
           cat "$GITHUB_WORKSPACE/src-overlay/conf/options.compat_linux" >> /usr/src/sys/conf/options
           cat "$GITHUB_WORKSPACE/src-overlay/conf/files.compat_linux" >> /usr/src/sys/conf/files.amd64
-          # COMPAT_LINUX must not already be live, or config(8) sees a duplicate.
-          grep -qE '^#COMPAT_LINUX' /usr/src/sys/conf/options.amd64
+          # We declare COMPAT_LINUX ourselves, so it must still be commented out
+          # upstream; LINPROCFS/LINSYSFS are already declared live in
+          # options.amd64 and must NOT be redeclared. Either assumption breaking
+          # on a FreeBSD bump is a "Duplicate option" failure in config(8), so
+          # assert both here where the message is legible.
+          grep -qE '^#COMPAT_LINUX[[:space:]]' /usr/src/sys/conf/options.amd64
+          grep -qE '^LINPROCFS[[:space:]]'     /usr/src/sys/conf/options.amd64
+          grep -qE '^LINSYSFS[[:space:]]'      /usr/src/sys/conf/options.amd64
           for s in amd64/linux/linux_sysvec.c amd64/linux/linux_genassym.c \
                    amd64/linux/linux_locore.asm amd64/linux/linux_vdso_gtod.c \
                    amd64/linux/linux_vdso.lds.s compat/linux/linux_vdso_inc.S \
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -204,7 +204,20 @@ jobs:
         run: |
           set -eu
           cat "$GITHUB_WORKSPACE/src-overlay/conf/options.compat_linux" >> /usr/src/sys/conf/options
-          cat "$GITHUB_WORKSPACE/src-overlay/conf/files.compat_linux" >> /usr/src/sys/conf/files.amd64
+          # The ABI itself is amd64-only for now (files.compat_linux references
+          # amd64/linux/* and x86/linux/*). arm64 has a complete Linux ABI in
+          # tree (sys/arm64/linux/) and mirroring this for it is a follow-up.
+          if [ "${{ matrix.target }}" = "amd64" ]; then
+            cat "$GITHUB_WORKSPACE/src-overlay/conf/files.compat_linux" >> /usr/src/sys/conf/files.amd64
+          else
+            # config/NEXTBSD is shared by both matrix legs, so every option it
+            # names must be DECLARED on both even where it selects nothing --
+            # otherwise config(8) fails with `unknown option "LINSYSFS"'.
+            # LINPROCFS/LINSYSFS live in options.amd64 upstream; COMPAT_LINUX we
+            # append to the MI sys/conf/options above, so it already covers both.
+            printf 'LINPROCFS\topt_dontuse.h\nLINSYSFS\topt_dontuse.h\n' \
+              >> /usr/src/sys/conf/options.arm64
+          fi
           # We declare COMPAT_LINUX ourselves, so it must still be commented out
           # upstream; LINPROCFS/LINSYSFS are already declared live in
           # options.amd64 and must NOT be redeclared. Either assumption breaking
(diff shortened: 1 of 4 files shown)


pkgdemon:
## CI validated — builds on both arches and boots

Four rounds of fixes; recording what each found since they were all real, and the last one is the interesting one.

| # | failed at | cause | fix |
|---|---|---|---|
| 1 | `config(8)`, <1s | `Duplicate option LINPROCFS` | `options.amd64:20-21` already declares LINPROCFS/LINSYSFS — only `COMPAT_LINUX` (line 18) is commented out. Dropped the redundant declarations. |
| 2 | `config(8)`, arm64 | `unknown option "LINSYSFS"` | `config/NEXTBSD` is shared by both matrix legs, so every option it names must be *declared* on both even where it selects nothing. Wiring is now arch-aware. |
| 3 | link, ~4 min | 6 undefined `kern_kmq_*` symbols | `linux_misc.c:2989-3076` calls the POSIX mqueue KPI unconditionally; `kern/uipc_mqueue.c` is `optional p1003_1b_mqueue`, which GENERIC doesn't set (mqueuefs.ko). Added `options P1003_1B_MQUEUE`. |
| 4 | — | not a failure | see below |

**Result:**

- `config(8)` accepted all five generated-artefact rules
- all 38 sources plus the `linux_genassym.o → linux_assym.h → linux_locore.o → linux_vdso.so.o` chain compiled with **zero warnings**
- amd64 **and** arm64 kernels linked
- the image **booted to the getty login prompt**, and the Mach/launchd/IOKit/configd suite passed

The thing I was least confident about — transcribing the vDSO's custom `ld --shared` and linker-script rules from `bsd.kmod.mk` into config(8) `dependency`/`compile-with` form — went through untouched.

## Unrelated finding: an expect race in the boot harness costs ~24 min

Worth its own issue on `nextbsd`, not a blocker here (the step is `continue-on-error: true` and the real gate passed).

The smoke-test ran ~25 minutes against a 2-6 minute baseline. Every on-image marker was actually emitted within half a second:

```
00:26:43.220  OK: boot reached the login prompt
00:26:43.285  MACH-SMOKE-OK
00:26:43.297  LIBSYSTEM-KERNEL-OK
00:26:43.306  MACH-PORT-OK
00:26:43.682  EVFILT-MACHPORT-em0: Using 1024 TX descriptors and 1024 RX descriptors   <-- spliced
00:26:43.773  EVFILT-MACHPORT-CONCURRENT-OK
00:26:43.784  TASK-SPECIAL-PORT-OK
00:34:43      WARN: EVFILT-MACHPORT marker not seen              (+8 min)
00:42:43      WARN: EVFILT-MACHPORT-CONCURRENT marker not seen   (+8 min)
00:50:43      FAIL: TASK-SPECIAL-PORT marker not seen            (+8 min)
```

`em0`'s attach printf spliced itself into the middle of `EVFILT-MACHPORT-OK`, so expect never matched it, stalled its full 480s, and by the time it advanced the next two markers had scrolled past — three cascading timeouts, ~24 minutes, all on markers that had already succeeded.

This is the hazard `nextbsd@165e56c` ("tests: unmute console") introduced: with the console un-muted, asynchronous kernel printfs interleave with the markers expect line-matches. Timing-dependent, so it will look intermittent. Two obvious mitigations: match markers with a leading anchor tolerant of mid-line splices, or drop the per-block timeout well below 480s for markers that are already `WARN`-on-timeout.

pkgdemon:
The ~24 min smoke-test overrun analysed above is filed separately as nextbsd-redux/nextbsd#405 — it is a pre-existing expect race in the boot harness (an `em0` attach printf splitting `EVFILT-MACHPORT-OK` in half), not caused by anything here, and non-gating. Noting it so the long CI time on this PR isn't mistaken for a regression it introduced.

PR #64, opened 2026-08-02
mach: don't deref a NULL ipc_space in ipc_entry_list_close — fixes panic on every Linux exit(2)

### LINKED ISSUE #63: kernel panics on first Linux exec — static Linuxulator from #61 takes the box down during linux_base-rl9 install
Regression from #61 (merged 2026-08-02 01:05Z). `main` currently ships a kernel that **panics on the first Linux `exec`**, so installing `emulators/linux_base-rl9` takes the machine down mid-install.

## Symptom

```
[2/3] Installing linux_base-rl9-9.8...
[2/3] Extracting linux_base-rl9-9.8: 100%
client_loop: send disconnect: Broken pipe
```

Machine reboots. Observed on real hardware (`system-version`), amd64.

## Why it only started with #61

`emulators/linux_base-rl9/pkg-plist.amd64` line 1 guards the install on the Linux ABI being present:

```
@preexec [ -n "`/sbin/sysctl -q %%SYSCTLMIB%%`" ] || ( echo "Cannot install package: kernel missing 64-bit Linux support" >&2; exit 1 )
```

with `SYSCTLMIB = compat.linux.osrelease`. Before #61 that sysctl did not exist, so pkg **refused the install cleanly**. #61 compiled the ABI in, the sysctl now exists, the guard passes, and pkg proceeds to line 2840:

```
@postexec %D/usr/sbin/ldconfig
```

`/compat/linux/usr/sbin/ldconfig` is a Linux ELF binary. That is the first genuine Linux `exec` on this kernel, and it panics. The timing in the transcript matches exactly — death immediately after extraction, during postexec.

So #61 did not break `pkg`; it removed the guard that was hiding an untested code path.

## Minimal reproducer

The payload is already on disk after a failed install, so no `pkg` needed:

```sh
/compat/linux/bin/true
```

## Already ruled out — build artifacts are correct

Both suspects I had going in are eliminated by inspecting the `kernel-obj-amd64` artifact from the merged build (run 30729461080):

| suspect | verdict | evidence |
|---|---|---|
| vDSO malformed / mis-embedded | **ruled out** | `linux_vdso.so.o` is a valid ELF 64-bit LSB shared object, x86-64, 5264 bytes. `linux_vdso_inc.o` carries `_binary_linux_vdso_so_o_start = 0` and `_binary_linux_vdso_so_o_end = 0x1490` — exactly 5264 bytes, so the `.incbin` picked up the right file at the right size. |
| `linux_genassym` offsets wrong | **ruled out** | Generated `linux_assym.h` reads `L_SC_R8=0, R9=0x8, R10=0x10, R11=0x18, R12=0x20, R13=0x28, R14=0x30, R15=0x38, RDI=0x40, RSI=0x48, RBP=0x50, RBX=0x58, RDX=0x60, RAX=0x68, RCX=0x70, RSP=0x78, RIP=0x80, RFLAGS=0x88, CS=0x90` — a clean 8-byte progression matching `struct l_sigcontext` field order, plus `LINUX_RT_SIGF_UC=0`, `LINUX_RT_SIGF_SC=0x28`, `LINUX_VERSION_CODE=0x50f00`. |

So the transcription of `sys/modules/linux64/Makefile` into config(8) form produced correct objects. The failure is a genuine static-vs-module runtime difference, not a bad build rule.

## Leading hypothesis: SYSINIT tie-break between linux_common and linux64elf

Both declare at **identical subsystem and order**:

```c
DECLARE_MODULE(linux_common, linux_common_mod, SI_SUB_EXEC, SI_ORDER_ANY);        /* linux_common.c:80  */
DECLARE_MODULE_TIED(linux64elf, linux64_elf_mod, SI_SUB_EXEC, SI_ORDER_ANY);      /* linux_sysvec.c:965 */
MODULE_DEPEND(linux64elf, linux_common, 1, 1, 1);                                 /* linux_sysvec.c:966 */
```

`MODULE_DEPEND` orders **`kldload`**. Compiled in, there is no loader — SYSINIT sorts by `(subsystem, order)` and these tie, leaving link order to decide. If `linux64elf` wins the tie, `linux64_elf_modevent` inserts the ELF brand entries before `linux_common_modevent`'s `MOD_LOAD` has run:

```c
	linux_dev_shm_create();
	linux_osd_jail_register();      /* the jail OSD slot linux_get_osname() reads */
	linux_netlink_register();
```

A Linux `exec` then reaches `linux_get_osname()`/`linux_get_osrelease()` against an unregistered OSD slot.

In the module world this cannot happen — `MODULE_DEPEND` forces `linux_common` first. It becomes reachable only when the ABI is compiled in, which is exactly what #61 does. **Hypothesis, not yet confirmed** — needs the backtrace.

If confirmed, the fix is small: give `linux_common` an explicit earlier `SI_ORDER` (or `linux64elf` a later one) via a patch in `patches/`, rather than relying on link order.

## Getting the backtrace

Panic output goes to the console, not SSH. In `/boot/loader.conf` (both are `CTLFLAG_RWTUN`, so they survive the reboot they prevent):

```
kern.panic_reboot_wait_time="-1"
debug.debugger_on_panic="1"
```

Then `/compat/linux/bin/true` on the console and `bt` at `db>`.

Interpretation guide:

- `linux_get_osname` / `linux_get_osrelease` / `osd_jail_get` → ordering hypothesis confirmed
- `linux_netlink_*` / `linux_dev_shm_*` → same family, same fix
- `linux_vdso_install` or exec image activation → vDSO wrong at runtime despite correct on-disk artifacts
- wild pointer in `linux_exec_imgact` / `elf64_*` → something else

## Exposure and mitigation

Anyone on current `main` who installs `linux_base-rl9` (directly, or as a dependency — e.g. `pkg install claude-code`) will panic and reboot, with the package left half-installed. Any retry re-panics.

Interim mitigation if this needs to stop being a footgun before the root cause is found: drop `options COMPAT_LINUX`, `LINPROCFS` and `LINSYSFS` from `config/NEXTBSD` while keeping `FUSEFS`. The two halves of #61 are independent — `FUSEFS` is what `fusefs-squashfuse` and the AppImage mount path need, and that path uses no Linux ABI at all. Removing `COMPAT_LINUX` also makes `compat.linux.osrelease` disappear, which restores the `@preexec` guard and makes pkg refuse cleanly again. The overlay fragments can stay in place, so re-enabling is three lines.

Refs #60, #61.

pkgdemon:
Fixes #63. Every Linux binary panicked the kernel **on exit(2)** — one line of real change.

## How it was found

The physical box gave nothing: instant reset, no panic text, no surviving msgbuf, and no netdump despite `igc` supporting debugnet. So I booted the CI-assembled image under qemu with a serial console and ran a 122-byte hand-built static Linux ELF (`_start: mov $60,%eax; xor %edi,%edi; syscall`). That produced the panic in full:

```
Fatal trap 12: page fault while in kernel mode
fault virtual address   = 0x28
fault code              = supervisor read data, page not present
instruction pointer     = 0x20:0xffffffff80f76a70
current process         = 48 (tiny_exit)
rbx: 0000000000000000
#3 0xffffffff810d9fcd at trap_pfault+0x37d
#4 0xffffffff810af9c8 at calltrap+0x8
#5 0xffffffff80b4a44c at exit1+0x33c
#6 0xffffffff811b45ae at linux_exit+0x2e
#7 0xffffffff810da606 at amd64_syscall+0x126
```

`nm` on the matching `kernel.full` resolves the faulting IP:

```
0xffffffff80f76a70 -> ipc_entry_list_close+0x170
```

So it is **our Mach layer**, reached through `exit1()`'s `process_exit` handler — not the Linuxulator.

## Why 0x28

`struct ipc_space` begins `is_ref_lock_data` (a `struct mtx`, 32 bytes), then `is_references` (4, padded to 8), then `is_entry_list`. That puts `is_entry_list` at **0x28** — exactly the fault address, and exactly what the trailing loop touches:

```c
	while(!LIST_EMPTY(&space->is_entry_list)) {
```

`space` was NULL.

## Why the existing guard missed it

`ipc_entry_list_close()` already carries an out-of-tree guard:

```c
	if (p->p_machdata == NULL || curthread->td_proc->p_machdata == NULL)
		return;
```

That is necessary but not sufficient. `p_machdata` being set does **not** imply the task owns an IPC space — `current_space()` is `current_task()->itk_space` (`sys/mach/thread.h:639`). A Linux-ABI process acquires a mach task lazily but never gets an `itk_space`, so it sails past the guard and hits the walk with `space == NULL`.

## The fix

Guard the space-owned walk rather than the whole function:

```c
	if (space == NULL)
		return;

	/* free unreferenced ipc_entrys */
```

Deliberately placed *after* the descriptor loops, not before them: a Linux process can still have inherited `DTYPE_MACH_IPC` descriptors that must be closed. Only the space-owned entry list has nothing to free. Behaviour for Mach processes is unchanged.

## This is not a Linuxulator bug

#61 didn't introduce it — it made a latent NULL deref *reachable*, by adding the first process ABI that doesn't allocate Mach IPC state. The field symptom was `pkg install linux_base-rl9` rebooting the machine mid-install: its `@postexec` runs `/compat/linux/usr/sbin/ldconfig`, and the panic fired when that process exited.

Worth noting the same guard gap would bite any future non-Mach ABI, not just Linux.

## Verification

CI build is the first gate. After that I'll re-run the qemu repro against the newly built image — same 122-byte binary, which should now exit cleanly with `RC=0` — and post the result here before merging.

### DIFF
--- a/src-overlay/sys/compat/mach/ipc/ipc_entry.c
+++ b/src-overlay/sys/compat/mach/ipc/ipc_entry.c
@@ -782,6 +782,21 @@ ipc_entry_list_close(void *arg __unused, struct proc *p)
 	FILEDESC_SUNLOCK(fdp);
 #endif
 
+	/*
+	 * Out-of-tree fix (nextbsd-kernel#63): p_machdata being set does NOT
+	 * imply the task owns an IPC space. A process whose ABI never allocates
+	 * one -- a Linux binary under COMPAT_LINUX acquires a mach task lazily
+	 * but no itk_space -- reaches here with current_space() == NULL, and the
+	 * walk below then reads NULL + offsetof(struct ipc_space, is_entry_list)
+	 * == 0x28 and panics in exit1(). Every Linux process died on exit(2).
+	 *
+	 * The descriptor loops above stay: such a process can still have
+	 * inherited DTYPE_MACH_IPC descriptors that must be closed. It is only
+	 * the space-owned entry list that has nothing to free.
+	 */
+	if (space == NULL)
+		return;
+
 	/* free unreferenced ipc_entrys */
 	i = 0;
 	while(!LIST_EMPTY(&space->is_entry_list)) {


pkgdemon:
## Verified on hardware-equivalent: fix confirmed

Booted the image CI assembled from this branch (`nextbsd-ci-image-amd64`, run 30731217116, kernel `20260802-035229`) under qemu and re-ran the exact repro that panicked the unfixed kernel.

```
[guest] sysctl -n compat.linux.osrelease; uname -r
5.15.0
20260802-035229
[guest] /tmp/tiny_exit; echo RC=$?
RC=0
*** LINUX EXEC SUCCEEDED ***
[guest] (/tmp/tiny_spin &) ; sleep 3; echo STILL_ALIVE
STILL_ALIVE
[guest] pkill -9 tiny_spin; echo KILLED
KILLED
```

**`Fatal trap` occurrences in the entire boot: 0.**

| case | unfixed kernel | this branch |
|---|---|---|
| `exit(0)` Linux binary | panic at `ipc_entry_list_close+0x170` | **RC=0** |
| Linux binary spinning in background | ran | **STILL_ALIVE** |
| `pkill -9` a live Linux process | **panic** — this is what reset the physical box | **KILLED, survived** |

Same binaries, same md5s (`2be3cdb3…`, `ad210e7f…`), same injection path as the failing run, so the only variable is the kernel.

The third row is the one that matters for the original report: on the physical box a no-syscall spin binary appeared to kill the machine, but the panic actually fired when I `pkill`'d it — the fault is on exit, not exec. That case now survives.

CI is green on both arches including the boot smoke-test. Merging.

PR #67, opened 2026-08-02
mach: skip fork bootstrap inheritance for tasks with no ipc_space — fixes AppImage panic

### LINKED ISSUE #66: mach: stale itk_bootstrap panics on fork() from a Linux process — ipc_port_copy_send via mach_task_fork_bsport
Second Mach NULL/stale-pointer panic of the same family as #63, exposed once #65 let claude-code get far enough to `fork()`.

**Not a regression from #61/#64/#65** — those made a previously unreachable path reachable. Filing rather than patching because the correct fix depends on the intended ownership model for `itk_bootstrap`, which is a judgement call for whoever owns the Mach layer.

## Panic

```
Fatal trap 12: page fault while in kernel mode
fault virtual address   = 0x490
fault code              = supervisor read data, page not present
instruction pointer     = 0x20:0xffffffff80b77819
current process         = 151 (claude)
rbx: fffff80014c19780   r13: fffff80014c19780
trap number             = 12
panic: page fault

#3 0xffffffff810dafcd at trap_pfault+0x37d
#4 0xffffffff810b0fa8 at calltrap+0x8
#5 0xffffffff80f8026a at ipc_port_copy_send+0x7a
#6 0xffffffff80f8a9b2 at mach_task_fork_bsport+0xf2
#7 0xffffffff80b522d8 at do_fork+0x10c8
```

Kernel `20260802-055026` (the ISO with #61/#62/#64/#65). Reproduced under qemu: install `claude-code`, mount the Linux ABI filesystems, run `claude` from any directory.

## Where

`src-overlay/sys/compat/mach/kern/task.c:1302`, in the `process_fork` handler:

```c
	if (IP_VALID(child_task->itk_bootstrap))
		ipc_port_release_send(child_task->itk_bootstrap);
	child_task->itk_bootstrap = ipc_port_copy_send(parent_task->itk_bootstrap);
```

The fault is inside `ipc_port_copy_send`, so it is `parent_task->itk_bootstrap` that is bad. That function early-returns on `!IP_VALID(port)`, so the pointer is neither `IP_NULL` nor `IP_DEAD` — it is non-NULL and points at memory that is not mapped. Fault address `0x490` is an offset within `struct ipc_port`, consistent with `ip_lock(port)` on a stale pointer.

The two guards above it both pass:

```c
	if (parent_task == NULL) return;
	if (parent_task->itk_bootstrap == IP_NULL) return;
```

## Hypothesis (unverified)

`ipc_entry_list_close` is registered for **both** `process_exit` and `process_exec` (`ipc_entry.c:813-814`) and tears down the IPC space. Grepping the overlay, **nothing anywhere clears or re-validates `itk_bootstrap`** — the only writes are the inheritance at `task.c:1302` and the release just above it.

That suggests: `sh` forks, the child inherits a send right into `itk_bootstrap`; the child execs `claude`, and the exec-time teardown destroys the port while `itk_bootstrap` keeps pointing at it; `claude` later forks and dereferences freed memory.

**What I cannot explain, and why this needs your eyes:** native processes exec-then-fork constantly and do not panic. So either the teardown spares the bootstrap port for processes that own an `ipc_space`, or Linux processes differ in a way that matters here — recall from #63 that a Linux process has `p_machdata` but a **NULL `itk_space`**, so `ipc_entry_list_close` returns early for it and its teardown path is genuinely different. I did not want to guess which invariant is the intended one.

## Possible directions

Listed for discussion, not as a recommendation:

1. Clear `itk_bootstrap` (with a release if valid) in the `process_exec` path, so a post-exec fork inherits nothing rather than a stale right.
2. Hold a real reference for the lifetime of `itk_bootstrap` so exec-time teardown cannot free it underneath the task.
3. Validate in `mach_task_fork_bsport` before copying — cheapest, but papers over the dangling pointer rather than fixing ownership.

## Impact

Any Linux binary that forks panics the machine. That is most real programs: claude-code does it during startup, and shells, package managers and build tools all fork. `claude --version` works (it does not fork); `claude` does not.

Same shape as #63 — the Mach layer assuming state that a non-Mach ABI process does not have — and worth an audit of the other `process_fork`/`process_exec`/`process_exit` handlers for the same assumption while this is being fixed.

Refs #60, #63, #65.

pkgdemon:
Fixes #66. Three-line guard; the rest is the explanation.

## Symptom

Any AppImage panics the machine within ~2 seconds of starting. `claude-code` survives, which is why this took a while to pin down.

## Backtrace

Reproduced under qemu on `NextBSD-amd64-20260802-055026` (all of #61/#62/#64/#65) with a serial console — hardware has no dump device, so a panic there just vanishes into a reboot.

```
Fatal trap 12: page fault while in kernel mode
fault virtual address   = 0xffffffff81239ce3
fault code              = supervisor write data, protection violation
current process         = 143 (thorium.AppImage)

#5  0xffffffff80c05bc8 at propagate_priority+0x58
#6  0xffffffff80c067d2 at turnstile_wait+0x2d2
#7  0xffffffff80b778c5 at __mtx_lock_sleep+0x175
#8  0xffffffff80f8ab53 at mach_task_fork_bsport+0x293
#9  0xffffffff80b522d8 at do_fork+0x10c8
#10 0xffffffff80b50c3e at fork1+0x83e
#11 0xffffffff811b49e8 at linux_fork+0x58
```

That is the `itk_lock(parent_task)` / `itk_lock(child_task)` pair. `itk_lock_data` is not an initialised `struct mtx`, so the lock is treated as contended, `turnstile_wait` enqueues against a bogus owner, and `propagate_priority` writes through a garbage pointer into kernel text.

## Two crash sites, one cause

An earlier run hit a **different** point in the same function:

| | earlier | this one |
|---|---|---|
| offset | `mach_task_fork_bsport+0xf2` | `mach_task_fork_bsport+0x293` |
| via | `ipc_port_copy_send+0x7a` | `__mtx_lock_sleep` → `turnstile_wait` → `propagate_priority` |
| fault | read `0x490`, page not present | **write** to kernel text, protection violation |

Two distinct failure modes in one function is the tell: guarding individual dereferences would be the wrong shape, because **the hook is operating on a task structure that is not valid for this process at all**. Whichever line runs first fails, depending on what the uninitialised memory happens to contain. That also explains claude-vs-AppImage without needing two bugs — same invalid state, different luck.

## The fix, and why this predicate

`p_machdata` being set does not mean the task is a real Mach task. #63/#64 already established the asymmetry: **a native process owns an `itk_space`; a Linux-ABI process acquires a task lazily and never does.** That is precisely why `ipc_entry_list_close` needed its guard in #64, and why `current_space()` (`current_task()->itk_space`) returns NULL for such a process.

So:

```c
	if (parent_task->itk_space == NULL)
		return;
```

placed before anything touches `itk_bootstrap` or takes `itk_lock`, since neither is valid to read *or* lock on a bogus task.

**Native behaviour is unchanged.** Those tasks have a space, so bootstrap-port inheritance for launchd-spawned daemons — the entire reason Task #39 added this hook, so `bootstrap_check_in` works for MachServices — still happens exactly as before. A Linux process has no space and cannot use Mach IPC at all, so there is nothing to inherit and skipping is correct rather than a workaround.

This is the third instance of the same class (#63/#64 on exit, this on fork), so an audit of the remaining `process_*` eventhandlers for the same assumption is probably worthwhile as a follow-up.

## Verification

I will boot the CI-built image from this branch and re-run the same AppImage before proposing a merge. The reproducer takes 2 seconds, so this is a real check rather than an inference — and given four of my automated checks tonight produced false passes by asserting on the absence of an error string, the criterion here is positive: the AppImage must run *and* the box must still be up afterwards.

### DIFF
--- a/src-overlay/sys/compat/mach/kern/task.c
+++ b/src-overlay/sys/compat/mach/kern/task.c
@@ -1251,6 +1251,35 @@ mach_task_fork_bsport(void *arg __unused, struct proc *p1, struct proc *p2,
 
 	if (parent_task == NULL)
 		return; /* parent has no Mach state — nothing to inherit */
+
+	/*
+	 * Out-of-tree fix (nextbsd-kernel#66): p_machdata being set does NOT mean
+	 * the task is a real Mach task. A Linux-ABI process acquires a task
+	 * lazily but never an itk_space — the same asymmetry #64 fixed in
+	 * ipc_entry_list_close, where current_space() is NULL for such a process.
+	 * Its itk_* fields are therefore not valid to read OR to lock, and this
+	 * hook died two different ways on them:
+	 *
+	 *   +0xf2  -> ipc_port_copy_send: read 0x490, page not present (stale
+	 *             itk_bootstrap port)
+	 *   +0x293 -> itk_lock -> __mtx_lock_sleep -> turnstile_wait ->
+	 *             propagate_priority: WRITE to kernel text, protection
+	 *             violation (itk_lock_data is not an initialised struct mtx,
+	 *             so the lock looks contended and the turnstile chain is
+	 *             garbage)
+	 *
+	 * Two distinct crash sites means guarding individual dereferences is the
+	 * wrong shape: the whole hook is operating on a structure that is not
+	 * valid for this process. Bail out instead.
+	 *
+	 * Native processes own a space, so bootstrap-port inheritance for
+	 * launchd-spawned daemons — the reason this hook exists (Task #39) — is
+	 * unaffected. A Linux process has no space and cannot use Mach IPC at
+	 * all, so there is genuinely nothing to inherit.
+	 */
+	if (parent_task->itk_space == NULL)
+		return;
+
 	if (parent_task->itk_bootstrap == IP_NULL)
 		return; /* parent has no bootstrap port — nothing to inherit */
 


pkgdemon:
## Verified against the workload that actually reproduces

Booted the CI image from this branch (run 30758109395) and ran the same Thorium AppImage that panics pre-fix, in the same environment (fresh overlay, no linux_base).

| | pre-fix (`NextBSD-amd64-20260802-055026`) | this branch |
|---|---|---|
| Thorium AppImage | **panic in ~2s** at `mach_task_fork_bsport+0x293` | **no panic** |
| `Fatal trap`/`panic:` in console | 1 | **0** |
| guest afterwards | dead | returned to a shell prompt |

The AppImage still fails, but for legitimate userland reasons on an image with no linux_base:

```
ELF interpreter /lib64/ld-linux-x86-64.so.2 not found, error 2
linux: jid 0 pid 143 (t.AppImage): unsupported umount2 flags 0x2
Error: No suitable fusermount binary found on the $PATH
Abort trap
```

Those messages matter: they prove the runtime **executed and forked** (it reached `umount2`), which is precisely the path that panicked before. A plain `Abort trap` is userland, not a kernel fault.

## A caveat worth recording: this is narrower than "fixes forking"

I first tried to verify with a 136-byte static Linux binary doing `fork()` + `exit(0)`. **It does not reproduce the bug** — I ran it against the UNFIXED published image as a control and it exited 0 cleanly, no panic. So a bare fork from a Linux process is not the trigger, and had I only run it against the fixed image I would have reported a meaningless pass.

That points at something this guard may not fully cover. `mach_task_fork_bsport` only initialises the child's mutexes on the `p2->p_machdata == NULL` path:

```c
	if (p2->p_machdata == NULL) {
		child_task = uma_zalloc(task_zone, M_NOWAIT | M_ZERO);
		...
		mach_mutex_init(&child_task->itk_lock_data, ...);
	} else {
		child_task = p2->p_machdata;   /* mutexes never initialised */
	}
```

A zeroed `struct mtx` is not `MTX_UNOWNED` (which is 4, not 0), so it reads as owned-by-thread-0 -> contended -> `turnstile_wait` -> `propagate_priority` walks a bogus owner. That is the `+0x293` backtrace exactly, and it suggests the deeper issue is a child that already carries a `p_machdata` whose locks were never initialised.

So: this PR verifiably closes the path AppImages take, and I would not claim more than that. #66 may deserve to stay open for the `else` branch above, or at least a follow-up. Your call on whether to close it on merge.

pkgdemon:
## Correction, then a proper controlled verification

My previous PASS on this PR **did not exercise the bug**. I ran the AppImage after `brandelf -t Linux`, but the original panic went through the *fallback brand* path (unbranded file + `kern.elf64.fallback_brand=3`). Those are different code paths in `__elfN(get_brandinfo)`, and only the fallback one panics.

Caught it by accident while verifying #68: a 136-byte static Linux binary doing `fork()`+`exit(0)` panics when unbranded but **not** when explicitly branded EI_OSABI=3. That also explains why my earlier minimal test came back clean against the unfixed image — it was the branded variant.

## Controlled A/B with a deterministic reproducer

Same binary, same `kern.elf64.fallback_brand=3`, only the kernel differs:

| image | has this fix | `fork_un` (EI_OSABI=0) | panics |
|---|---|---|---|
| PR #68 branch (brand fix only) | no | **panic** `mach_task_fork_bsport+0xf2` -> `linux_fork+0x58` | 1 |
| **this branch** | yes | **UNRC=0**, clean exit | **0** |

Console panic count is 0 across the entire boot on this branch.

The AppImage run (native-root mounts, no linux_base) also completes without a panic and fails only in userland — `ELF interpreter /lib64/ld-linux-x86-64.so.2 not found`, `No suitable fusermount binary`, `Abort trap` — having demonstrably forked first (it reaches `umount2`).

## Reproducer for anyone revisiting #66

136 bytes, no packages, no network, no linux_base:

```
_start: mov $57,%eax; syscall      # fork
        mov $60,%eax; xor %edi,%edi; syscall   # exit(0)
```

built with **EI_OSABI = 0**, then `sysctl kern.elf64.fallback_brand=3` and run. Panics in ~2s on an unfixed kernel. The branded variant does not, which is worth knowing before writing a test.

## Still not claiming this closes all of #66

As noted earlier, `mach_task_fork_bsport` reuses an existing `p2->p_machdata` without initialising its mutexes, and a zeroed `struct mtx` is not `MTX_UNOWNED`. This guard closes the path Linux processes take; the `else` branch may still be reachable another way. Suggest #66 stays open or gets a follow-up rather than being auto-closed.

PR #68, opened 2026-08-02
kernel: default elf64 fallback_brand to ELFOSABI_LINUX so unbranded Linux binaries (AppImages) can exec

pkgdemon:
Refs #60, nextbsd-userland#54. `patches/0009`, added to `patches/series`, verified to apply cleanly against `releng/15.1`.

Separate from #67 on purpose: that one is a panic fix that should merge on its own merit, this one is a **policy change** and deserves its own discussion.

## Why it is needed

A stock AppImage cannot execute at all:

```
$ ./Thorium_Browser_150.0.7871.101_AVX.AppImage
ELF binary type "0" not known.
-sh: ./Thorium_...AppImage: Exec format error
```

FreeBSD resolves an ELF's ABI in this order — and an AppImage defeats the first three:

| step | AppImage |
|---|---|
| `EI_OSABI` | `UNIX - System V` (0) — no help |
| `.note.ABI-tag` | absent; only `.note.gnu.build-id` |
| interpreter path | **no `PT_INTERP` at all** — the type2 runtime is static-pie musl |
| `kern.elf64.fallback_brand` | `-1` → give up |

Verified with `readelf` on the actual file. So the fallback is the only remaining hook, and this patch changes its default from `-1` to `3` (`ELFOSABI_LINUX`).

## Why `linux_base` works today and this is still needed

`x11/linux_base-rl9` brands its loader at build time:

```make
	${BRANDELF} -t Linux ${WRKSRC}/usr/lib64/ld-linux-*.so.2
```

so every **dynamically** linked Linux binary is identified by its interpreter path. Static ones have no such hint. That is exactly why `/compat/linux/bin/sh` runs fine while a downloaded AppImage does not.

`brandelf -t Linux` per file is not a workable alternative for AppImages, since users download them and it modifies the file.

## Why the kernel default

`rc.d/linux` sets this sysctl. NextBSD has launchd as PID 1, no `rc.d`, and never reads `/etc/sysctl.conf` — so as with `0008` (`vfs.usermount`), the kernel default is the only mechanism that persists. It remains `CTLFLAG_RWTUN` upstream, so `kern.elf64.fallback_brand=-1` in `/boot/loader.conf` restores the stock posture.

## 32-bit deliberately left alone

`imgact_elf.c` is compiled twice via `__elfN()`, so a naive `= -1` → `= 3` would change **both** ABIs. Guarded on `__ELF_WORD_SIZE == 64`, because #61 compiled in only the 64-bit Linux ABI — branding i386 ELFs as Linux would point them at an ABI that is not present, and a clear `Exec format error` beats failing deeper in.

## The tradeoff, stated plainly

This is broader than the mount work. It makes **any** unbranded ELF64 on the system be treated as Linux, including something that is not Linux at all and would otherwise have failed cleanly at exec. I raised this on nextbsd-userland#54 and my preference there was to gate it on a Linux userland actually being installed, via the same trigger that mounts the filesystems — which keeps the default posture unchanged on images with no Linux runtime.

You asked for it baked into the kernel, so that is what this does. If you would rather have the gated version instead, the alternative is a `sysctl` call in the linuxfs job in nextbsd-userland#53 and this patch gets dropped — happy to switch.

## Verification

Not yet booted. Will confirm against the CI image from this branch that an unbranded static Linux binary execs without `brandelf`, before proposing a merge.

### DIFF
--- a/patches/0009-NextBSD-default-elf64-fallback_brand-to-ELFOSABI_LINU.patch
+++ b/patches/0009-NextBSD-default-elf64-fallback_brand-to-ELFOSABI_LINU.patch
@@ -1,0 +1,82 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: pkgdemon <jpm820@gmail.com>
+Date: Sun, 2 Aug 2026 12:20:00 -0500
+Subject: [PATCH] NextBSD: default elf64 fallback_brand to ELFOSABI_LINUX
+
+Linux binaries are overwhelmingly built with EI_OSABI = ELFOSABI_NONE (0).
+FreeBSD identifies an ELF's ABI by, in order: EI_OSABI, a .note.ABI-tag, the
+interpreter path, then kern.elf64.fallback_brand. A STATICALLY linked Linux
+binary defeats the first three -- no OSABI, often no ABI note, and no PT_INTERP
+segment at all -- so it reaches the fallback. At -1 the loader gives up:
+
+	ELF binary type "0" not known.
+	Exec format error
+
+AppImages are exactly this case; the type2 runtime is static-pie musl (verified
+with readelf: OS/ABI "UNIX - System V", no INTERP, only .note.gnu.build-id).
+Because a user downloads them, brandelf(1) per file is not a workable
+alternative.
+
+x11/linux_base-rl9 works today only because the port brandelfs its loader
+(BRANDELF -t Linux .../ld-linux-x86-64.so.2), so DYNAMIC Linux binaries are
+identified by interpreter path. Static ones have no such hint.
+
+rc.d/linux sets this sysctl for the same reason. NextBSD has launchd as PID 1
+and no rc.d, and /etc/sysctl.conf is never read, so the kernel default is the
+only persistent mechanism -- the same reasoning as patch 0008 (vfs.usermount).
+
+32-bit is deliberately left at -1: nextbsd-kernel#61 compiled in only the 64-bit
+Linux ABI, so branding i386 ELFs as Linux would point them at an ABI that is not
+present. A clear "Exec format error" beats failing deeper in.
+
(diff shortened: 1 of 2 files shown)


pkgdemon:
## Verified

Booted the CI image from this branch (run 30758153786, kernel `20260802-171231`).

```
sysctl -n kern.elf64.fallback_brand   ->  3        (fresh boot, no sysctl set by hand)
```

And an **unbranded** 136-byte static Linux binary (`EI_OSABI = 0`, confirmed by `brandelf`: "is of brand 'SVR4' (0)") **executes** — it reaches `linux_fork`, which it could only do having been branded as Linux by the fallback. On a pre-#68 kernel the same file gives `ELF binary type "0" not known / Exec format error`.

Positive evidence rather than absence-of-error: the binary ran far enough to enter the Linux fork path.

(On this branch it then panics in `mach_task_fork_bsport`, because this branch does not carry #67. That is expected and is fixed by #67, verified separately there — and it incidentally gave me the deterministic reproducer for #66.)

32-bit left at `-1` as designed.

pkgdemon:
Same note as on #62, for the record only: 'NextBSD ... never reads `/etc/sysctl.conf`' is not accurate -- `launchctl bootstrap` applies the file (`nextbsd-userland support/launchctl.c:2326`, `apply_sysctls_from_file`, `:4330-4365`), so `kern.elf64.fallback_brand=3` could also have lived there. The kernel default is still the better choice for the reasons the PR gives (reaches installs on upgrade, `RWTUN` loader.conf override, and the overlay that owns `/etc` is seed-once). Runtime application of sysctl.conf is unverified on a running box. E15 B2 (parked) keeps rc.d/linux's 'set only if -1' guard so a `loader.conf` override is never clobbered.

_Tracked under nextbsd/nextbsd#471 (E15). Plan: https://pkgdemon.github.io/nextbsd-diskarbitration-mounts-plan.html#corrections_

PR #78, opened 2026-08-23
ci: emit kernel.bin (arm64 booti image) for the Pi 500+ bring-up

pkgdemon:
Unblocks phase 1 of the [Pi 500+ native-boot plan](https://pkgdemon.github.io/nextbsd-rpi500-native-boot-plan.html) — see the [image sub-plan](https://pkgdemon.github.io/nextbsd-rpi500-image-plan.html) for the full context.

## Why

The Pi 500+ has no UEFI (worproject/rpi5-uefi is archived, targets pre-D0 `BCM2712C1`, and never added board type `0x19`), so NextBSD boots there by having the Pi firmware load the kernel **directly**. That requires the kernel wrapped in the arm64 Linux `Image` header — the firmware parses it, copies the payload to `0x200000`, and enters at EL2 with `x0` = physical FDT.

FreeBSD already produces exactly this, and nothing in this repo invoked it. That one missing make target was the whole distance between the project and a FreeBSD banner on real hardware.

| Piece | Location | Status |
|---|---|---|
| `kernel.bin` target | `sys/conf/Makefile.arm64` | already there |
| `write_v8booti()` | `sys/tools/arm_kernel_boothdr.awk` | already there |
| `LINUX_BOOT_ABI` | `sys/arm64/conf/std.arm64` | already in GENERIC |

## What this does

One step after `buildkernel`, arm64 only, using the same `buildenv BUILDENV_SHELL=` pattern the existing `sysent` step uses. It belongs in this repo because the boothdr awk script and the cross `objcopy` both live in this container.

It **asserts the magic at offset 56** (`41 52 4d 64`) rather than trusting the exit code. A silently headerless image would boot to a black screen on the board, where a missing header is indistinguishable from a hang — so it fails CI instead.

`kernel.bin` lands inside the `sys/NEXTBSD` obj subtree, so the existing tar picks it up with no packaging change. It is also uploaded on its own as `rpi500-kernel8-arm64` so the bring-up can fetch just that file rather than unpacking the obj tree.

amd64 is untouched.

## Testing

CI proves the header exists. It cannot prove the board boots — nothing in CI can boot a Pi 500+. The real test is writing it to a USB stick with `config.txt` + `bcm2712-rpi-500.dtb` and watching the serial console, which is now live on the hardware (phase 0 complete: firmware handoff → kernel messages → `login:` over the 40-pin header at 115200).

**Not for merge until it has booted the board.**


probonopd:
Let me know  if I can help testing. I have a Pi 500 8GB and a Pi 5 4GB.

PR #79, opened 2026-08-23
pci: bring up the BCM2712 host bridge on the Pi 500+

pkgdemon:
Phase 3 of the [Pi 500+ native-boot plan](https://pkgdemon.github.io/nextbsd-rpi500-native-boot-plan.html). Follows #78, which got the board booting to `mountroot`.

## Why

At `mountroot>`, `?` lists **zero** GEOM disks. That is correct and expected — every disk and both USB controllers are behind PCIe:

```
BCM2712
 ├─ pcie1 → NVMe 1d97:5216           ← the disk
 └─ pcie2 → RP1 1de4:0001 → dwc3 → xHCI → USB-HID   ← the keyboard
```

## What it is

Smaller than the plan assumed. FreeBSD already ships `bcm2838_pci` for the previous generation of the same Broadcom IP, already subclassing `generic_pcie_fdt_driver` with link training and PERST handling. OpenBSD's `bcmpcie(4)` covers 2711 and 2712 **in one file**, so the generational delta is readable directly rather than reverse-engineered — and it is ISC, unlike Linux's GPL-2.0-only equivalent, which matters for the vendoring rule.

| Delta | |
|---|---|
| `PCIE_HARD_DEBUG` | `0x4204` → `0x4304`, so the offset becomes per-SoC state |
| refclk | 54 MHz clock selected through the SerDes PHY over MDIO — a capability this driver lacked |
| read errors | return `0xffffffff`, not `0xdeaddead` |
| L1SS | PM clock period 18 |

SoC identified at runtime from the compatible string, so no new `opt_soc.h` dependency in the driver.

## Deliberately not done

- **Inbound DMA window is not programmed on 2712.** The hardcoded 2711 values would be actively wrong — 2712's is a non-identity 64 GiB aperture mapping PCIe `0x10_0000_0000` → CPU `0x0` — and FreeBSD has no `dma-ranges` parsing anywhere ([D58105](https://reviews.freebsd.org/D58105), still in review). Enumeration does not need it; bus-mastering does.
- **MSI untouched.** 2712 routes it through a separate MIP block remapping to GIC SPIs.

## Testing

CI can only prove it builds. The milestone is one boot on the board: **`pcib` attaches, the link trains, and RP1 + the NVMe appear.** Not for merge until that is measured.

### DIFF
--- a/config/NEXTBSD-RPI500
+++ b/config/NEXTBSD-RPI500
@@ -59,3 +59,16 @@ env		"rpi500.env"
 # the Pi 4's.
 nooptions	SOC_BRCM_BCM2837
 nooptions	SOC_BRCM_BCM2838
+
+# ...but do enable this board's own SoC, which gates the PCIe host bridge.
+#
+# Everything useful on a Pi 500+ is behind PCIe: the NVMe on pcie1, and RP1 --
+# and therefore both USB controllers, and therefore the keyboard -- on pcie2.
+# Without this the kernel reaches mountroot and `?` lists zero disks.
+#
+# The driver is FreeBSD's own bcm2838_pci, extended to match
+# "brcm,bcm2712-pcie" (patches/0013). It has to be compiled in rather than
+# shipped as a kext: route (a) has no loader, so nothing loads a .ko at boot,
+# and a kext delivering PCIe would have to come off a disk that only exists
+# once PCIe works.
+options 	SOC_BRCM_BCM2712
--- a/patches/0013-bcm2838_pci-support-the-BCM2712-PCIe-controller.patch
+++ b/patches/0013-bcm2838_pci-support-the-BCM2712-PCIe-controller.patch
@@ -64,7 +64,7 @@ block that remaps to GIC SPIs.
  arm/broadcom/bcm2835/raspberrypi_virtgpio.c		optional soc_brcm_bcm2837 gpio fdt | soc_brcm_bcm2838 gpio fdt
 --- a/sys/arm/broadcom/bcm2835/bcm2838_pci.c
 +++ b/sys/arm/broadcom/bcm2835/bcm2838_pci.c
-@@ -61,6 +61,31 @@
+@@ -61,6 +61,39 @@
  #define BRIDGE_DISABLE_FLAG	0x1
  #define BRIDGE_RESET_FLAG	0x2
  #define REG_PCIE_HARD_DEBUG			0x4204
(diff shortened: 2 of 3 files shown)


pkgdemon:
Measured on the board. **PCIe is up and the NVMe attaches.**

```
pcib0: <BCM2838/BCM2712-compatible PCI-express controller> mem 0x1000110000-0x100011930f irq 50,51 on simplebus1
pcib0: hardware identifies as revision 0x500.
pci0: <OFW PCI bus> on pcib0
pcib1: <PCI-PCI bridge> irq 72 at device 0.0 on pci0
pci1: <PCI bus> on pcib1
nvme0: <Generic NVMe Device> at device 0.0 on pci1
nvme0: Allocated 64MB host memory buffer
nda0 at nvme0 bus 0 scbus0 target 0 lun 1
nda0: nvme version 1.4
```

And at the `mountroot>` prompt, `?` now lists real storage where it previously listed nothing:

```
nda0  nda0s1  nda0s2  msdosfs/bootfs  ext2fs/rootfs  diskid/DISK-QDT2014102964
```

GEOM tasted the partition table and both filesystems by label, so block reads — and therefore DMA — work.

## Two bugs found on hardware, both in this PR

1. **MISC_CTRL.** `REG_DMA_CONFIG` (0x4008) is not a DMA register; it is MISC_CTRL, and `DMA_WINDOW_ENABLE` (0x3000) is `SCB_ACCESS_EN | CFG_READ_UR_MODE`. Skipping the write to avoid the wrong 2711 window values also cleared those, so the controller never left reset.
2. **PERST#.** `BRIDGE_DISABLE_FLAG` is a misleading name for `PCIE_RGR1_SW_INIT_1_PERST`. 2712 moved that bit to `PERSTB` in `REG_MISC_PCIE_CTRL` (0x4064), with inverted sense, so the device was held in reset and the link never trained.

Both surfaced identically as `controller failed to start`.

## Still outstanding

- **MSI**: `nvme0: unable to allocate MSI-X` / `unable to allocate MSI`, as expected — 2712 routes MSI through the separate MIP block, not yet ported. The device works without it.
- **Inbound DMA window** still not programmed. It evidently works for what has been exercised so far; whether it holds for allocations above the aperture is untested and still wants `dma-ranges` (D58105).
- **pcie2 (RP1) is masked off** by a `nopcie2` device-tree overlay during bring-up, because RP1 is where the serial console lives and resetting that link makes the board go silent. That is a scaffold, not part of this PR.

Issue #81, opened 2026-08-23
rpi500: port the RP1 bridge (rpone) — unblocks USB, keyboard and ethernet


pkgdemon:
Phase 4 of the [native-boot plan](https://pkgdemon.github.io/nextbsd-rpi500-native-boot-plan.html#seq). **Blocked by #80.**

RP1 is the southbridge behind `pcie2`, and everything the machine needs to be usable hangs off it:

```
pcie2 -> RP1 1de4:0001
          |- 1f00100000 ethernet [raspberrypi,rp1-gem, cdns,macb]
          |- 1f00200000 usb [snps,dwc3] -> xhci-hcd.0
          `- 1f00300000 usb [snps,dwc3] -> xhci-hcd.1
                         `- 3-2 2e8a:0011 "Raspberry Pi Ltd Pi 500+ Keyboard (ANSI)"
```

RP1 is a PCIe endpoint that exposes an entire simple-bus of peripherals behind a BAR. OpenBSD models it in `sys/arch/arm64/dev/rpone.c` — **219 lines, ISC** — by having the bridge inherit `simplebus`, so the PCIe node doubles as an FDT bus for RP1's subnodes. The plan calls that trick worth copying verbatim.

## Why it is blocked

Bringing up `pcie2` means resetting the link RP1 is on, and RP1 carries the serial console. Today that makes the board go silent with no way to tell a working driver from a dead one. `pcie2` is currently masked off by a `nopcie2` overlay for exactly this reason.

## State

`pcie1` already works — the host bridge, PCI enumeration and NVMe all attach (#79). The controller driver is generation-agnostic, so `pcie2` should need no new controller work; this issue is the RP1 bridge on top of it.

pkgdemon:
## Design note — how RP1 should attach on FreeBSD

Measured the shape of the problem on the board before writing anything.

**RP1 is `compatible: simple-bus`**, and its peripherals are device-tree children of the PCIe node, not PCI functions:

```
pcie@1000120000
  rp1  (simple-bus)
    ethernet@100000  usb@200000  usb@300000  gpio@d0000  i2c@70000  dma@188000 ...
```

So the driver has to bridge two worlds: a PCI endpoint on one side, an FDT bus on the other. OpenBSD solves it by having `bcmpcie` inherit `simplebus_softc` so the PCIe node doubles as an FDT bus.

[...]

Issue #87, opened 2026-08-23
Write a Claude skill for Raspberry Pi board bring-ups


pkgdemon:
Bringing NextBSD up on the Raspberry Pi 500+ produced a lot of reusable procedure, and almost none of it is written down anywhere an agent would find it. It currently lives in commit messages, two pages on pkgdemon.github.io, and a chat log. The next board — a Pi 5 B, a CM5, or the next generation — would start from scratch.

Propose a `.claude/skills/rpi-bringup/` skill that encodes it.

## Why a skill rather than a doc

The material is *procedural* and mostly about **how to make a silent board legible**. Everything on this hardware fails the same way: no output. A kernel that never started, a kernel that crashed in `locore`, a kernel running fine with no console, a driver that reset the controller the console lived behind, and an MSI routed to the wrong SPI all present identically — as silence. Most of the weekend went on building instruments to tell those apart, and that is exactly the kind of knowledge an agent needs *before* it starts guessing.

## What it should encode

**The unattended test harness.** This is the highest-value part.
- `tryboot` as a one-shot boot: `reboot '0 tryboot'` reads `tryboot.txt` instead of `config.txt`, and any failure reverts on the next power cycle. `config.txt` is never touched, so the fallback is structural.
- `BOOT_ORDER` so the default boot is always the working OS. `0xf461` = SD → NVMe → USB; nibbles read LSB-first.
- **Recovering a board sitting at `mountroot>` without physical access**: send a newline, manual input aborts, the mount genuinely fails, the kernel panics and auto-reboots, `tryboot` is spent, and ssh comes back. This is what makes the loop autonomous.
- The limit: a kernel that *hangs* rather than resets still needs a physical power cycle.

**Reading the console from the host.** The Debug Probe is a USB device on the *Mac*, so it stays enumerated no matter what the board does — no reconnect logic needed. macOS resets the baud when the fd closes, so a logger must hold it open: `exec 3<>/dev/cu.usbmodem*; stty -f <dev> 115200 raw -echo; cat <&3`. Only one reader at a time; `screen` must be quit first.

**Which UART, and the trap.** Two consoles exist on a Pi 5-family board and the choice is load-bearing:
- 40-pin header = RP1 UART0, behind PCIe. No disassembly, but **it dies the instant a driver owns that controller**, and a working driver then looks exactly like a dead one.
[...]

PR #91, opened 2026-08-24
rp1: interrupt controller over RP1's MSI-X vectors

pkgdemon:
Second increment on #81. Enumeration landed in #89; this makes RP1 a PIC so its children can actually allocate an IRQ.

Without it `xhci` cannot attach behind either `snps_dwc3_fdt` controller — the resource lists are built from DT interrupts, and there is nothing to map them onto.

## The mapping is 1:1, not demultiplexed

RP1 internal IRQ N **is** MSI-X vector N. There is no status register to read in the handler: the vector is the identity. Per-vector configuration lives in BAR1 at `0x108008 + 4*N`, with set/clear aliases at `+0x800`/`+0xc00`.

So the driver claims all 61 vectors, registers an `intr_irqsrc` for each, becomes a PIC on the `rp1` device-tree node, and hooks a filter per vector that dispatches its own source.

Level-triggered sources set `IACK_EN` so the block acknowledges on read; edge-triggered ones must not, or the interrupt is lost. That distinction comes from the second FDT interrupt cell, and `post_filter` acknowledges only for level.

Ordering matters: the controller is stood up **before** children are added, because their resource lists cannot be mapped until this node is a registered PIC.

Register layout and `IACK_EN` semantics are from OpenBSD's `rpone(4)`, which is ISC.

## Note on the patch

The first push of this branch shipped a malformed `0019` — header plus *two* conflicting diffs from the same base, because the regeneration spliced on `^---$` and the previous file had no such line, so `sed` emitted the whole thing including its diff. Repaired in a separate commit; the series is now verified by applying `0018` then `0019` to a pristine tree and inspecting the result rather than trusting an exit status.

Refs #81.


pkgdemon:
Board tested, and it failed — usefully.

```
pcib2: <BCM2838/BCM2712-compatible PCI-express controller> mem 0x1000120000-0x100012930f irq 52,53 on simplebus1
pci2: <OFW PCI bus> on pcib2
pcib3: <PCI-PCI bridge> irq 78 at device 0.0 on pci2
pci3: <PCI bus> on pcib3
rp10: <RP1 southbridge> at device 0.0 on pci3
rp10: could not allocate MSI-X vectors
device_attach: rp10 attach returned 6
```

That message reads like the MSI controller declining to hand out vectors. It is not — it is `pci_alloc_msix()` returning ENXIO before it ever reaches the MSI controller:

```c
/* Make sure the appropriate BARs are mapped. */
rle = resource_list_find(&dinfo->resources, SYS_RES_MEMORY,
    cfg->msix.msix_table_bar);
if (rle == NULL || rle->res == NULL ||
    !(rman_get_flags(rle->res) & RF_ACTIVE))
        return (ENXIO);
```

Read off the endpoint rather than guessed at:

```
Region 0: Memory at 1f00410000 (32-bit, non-prefetchable) [size=16K]
Region 1: Memory at 1f00000000 (32-bit, non-prefetchable) [size=4M]
Capabilities: [b0] MSI-X: Enable+ Count=61 Masked-
        Vector table: BAR=0 offset=00000000
        PBA:          BAR=0 offset=00002000
```

**Table and PBA are both in BAR0**, and the driver mapped only BAR1. BAR1 is the window every peripheral lives in and the only thing this driver reads or writes; nothing here ever dereferences BAR0, because the PCI layer writes the table on our behalf. So the omission was invisible until the hardware refused — and then refused in the vocabulary of a different subsystem.

Fixed in the commit above. `Count=61` also confirms `RP1_MSIX_VECTORS` from the silicon rather than from the OpenBSD driver the layout came from.

## Two things worth recording separately

**The harness closed the loop with no power cycle.** `tryboot` → NextBSD → `mountroot>` → bare CR → manual input aborts → mount fails → panic → reboot → `tryboot` spent → Raspberry Pi OS → ssh answers. Twice, unattended. That is the procedure #87 encodes, working as designed.

**`-v` never took effect.** `nextbsd-cmdline.txt` holds `FreeBSD: -v` and the boot is not verbose. `parse_fdt_bootargs()` only parses when `fdt_get_chosen_bootargs()` succeeds, so the firmware appears not to be setting `/chosen/bootargs` under `tryboot` at all. Not blocking here — the failure was diagnosable from the endpoint itself — but it removes an instrument, and I would rather not rediscover it later. Filing separately.

pkgdemon:
Board tested with the BAR0 fix. **The interrupt controller works.**

```
rp10: <RP1 southbridge> at device 0.0 on pci3
snps_dwc3_fdt0: <Synopsys Designware DWC3> mem 0xc040200000-0xc0402fffff irq 172 on rp10
snps_dwc3_fdt0: Failed to map memory
device_attach: snps_dwc3_fdt0 attach returned 6
snps_dwc3_fdt0: <Synopsys Designware DWC3> mem 0xc040300000-0xc0403fffff irq 173 on rp10
snps_dwc3_fdt0: Failed to map memory
device_attach: snps_dwc3_fdt0 attach returned 6
mmio_sram0: <MMIO SRAM> mem 0xc040400000-0xc04040ffff on rp10
mmio_sram0: Can't allocate resources for device.
```

RP1 attaches — no MSI-X error — and **`irq 172` and `irq 173` were mapped and handed to children**. That is the whole point of this PR: children could not allocate an IRQ at all before, and now they can. Both DWC3 controllers are found, which is the enumeration this unblocks.

## What fails now is a different thing

Memory, not interrupts. `0xc040200000` is the child address passed through untranslated.

Read off the board:

```
rp1:  #address-cells = 2, #size-cells = 2
pcie: #address-cells = 3      <- PCI addressing, 3 cells
rp1/ranges:      child 0xc0_4000_0000  ->  PCI mem 0x0,  size 0x410000
usb@200000/reg:        0xc0_4020_0000,     size 0x10_0000
```

The rp1 node's parent is a **PCI** node, so its `ranges` names the parent in PCI address form — three cells, the first being `phys.hi` (`0x02000000`, 32-bit memory space). `simplebus_fill_ranges()` assumes a plain parent address and cannot translate that, so it passes the child address through unchanged, and `0xc040200000` is what a child then tries to map.

The correct target is `rman_get_start(sc->bar) + parent_offset` — i.e. BAR1's assigned base plus `0x200000` for `usb@200000`. The single range covers `0x410000`, which is BAR1 (4 MB) plus BAR2 (64 KB) laid end to end in the device's own view.

That is a defect in the **bridge** half (the enumeration from #89), newly visible only because we now get far enough to see it, so it is not in scope here. Filing it and doing it as the next increment — it is the last structural piece before `xhci` attaches and the built-in keyboard can enumerate.

## Harness note

Three unattended board tests tonight, zero power cycles.

Issue #98, opened 2026-08-24
rp1: no Ethernet driver — cgem(4) does not match raspberrypi,rp1-gem / cdns,macb


pkgdemon:
Milestone 1 requires working Ethernet. There is no driver for this MAC.

Measured on the board:

```
eth0 -> /sys/devices/platform/axi/1000120000.pcie/1f00100000.ethernet
compatible = "raspberrypi,rp1-gem", "cdns,macb"
Linux driver: macb
```

It is an RP1 child at BAR1 `+0x100000`, so it is behind the same address-translation bug as USB (#95) — but unlike USB, a correct translation is **not** expected to be sufficient, because nothing in the tree binds to it.

## Why cgem(4) does not just work

FreeBSD has a Cadence GEM driver, and `raspberrypi,rp1-gem` is a Cadence GEM. But `sys/dev/cadence/if_cgem.c` matches only:

```c
{ "cdns,zynq-gem",         HWQUIRK_RXHANGWAR },
{ "cdns,zynqmp-gem",       HWQUIRK_NEEDNULLQS },
{ "xlnx,zynq-gem",         HWQUIRK_RXHANGWAR },
{ "xlnx,zynqmp-gem",       HWQUIRK_NEEDNULLQS },
{ "microchip,mpfs-mss-gem", HWQUIRK_NEEDNULLQS },
{ "sifive,fu540-c000-gem", HWQUIRK_NONE },
{ "sifive,fu740-c000-gem", HWQUIRK_NONE },
```

No `cdns,macb`, no `raspberrypi,rp1-gem`.

## Approach

Start by adding a compat entry and seeing how far it gets — that is one line and answers the interesting question immediately. Expect quirk work: the quirk column above exists precisely because these blocks differ in DMA queue handling and RX hang behaviour, and RP1's variant has never been driven here.

Note also that Linux uses `macb`, which covers both MACB and GEM generations; `cgem(4)` targets GEM. Confirm which the RP1 block actually is before assuming register compatibility.

Also needed: the PHY, and the MAC address, which the firmware patches into the device tree (`smsc95xx.macaddr=88:A2:9E:49:6E:B3` appears in `/chosen/bootargs`, and the DT node carries one too).

Blocked by #95. Part of the milestone-1 gate.

pkgdemon:
Unblocked by the #95 fix. The node is now reachable and correctly translated:

```
[...]

PR #103, opened 2026-08-24
cgem: match RP1's Cadence GEM, and take the MAC from the device tree

pkgdemon:
Two patches toward #98. **Board-tested** — the driver attaches. It does not yet have link; see below for exactly what remains.

## 1. Match `raspberrypi,rp1-gem`

`cgem(4)` is already compiled in on arm64 (GENERIC includes `std.xilinx`, which carries `device cgem`), so this was only a matching question.

`HWQUIRK_NONE` rather than `HWQUIRK_NEEDNULLQS`, chosen on evidence rather than caution. RP1's `macb_config` in the vendor kernel differs from `zynqmp_config` in exactly two ways: it adds `MACB_CAPS_CLK_HW_CHG` and omits `MACB_CAPS_BD_RD_PREFETCH`. Descriptor prefetch is *why* the Zynq parts need the null queue pointers programmed, so a part without it should not need them.

Result on the board:

```
cgem0: <Cadence CGEM Gigabit Ethernet Interface> mem 0xc040100000-0xc040103fff irq 166 on rp10
```

It attaches. The quirk call looks right.

## 2. Take the MAC address from the device tree

```
cgem0: no mac address found, assigning random: 62:73:64:bd:40:7c
```

`cgem_get_mac()` reads only the controller's address registers, assuming a boot loader filled them in. Route (a) has no boot loader. The real address is in the device tree, where the firmware patches it:

```
ethernet@100000/local-mac-address = 88 a2 9e 49 6e b3
```

— the same address Linux uses for `eth0` on this board. Now checked first, falling back to the registers and then to random. A multicast or all-zero property is treated as absent rather than trusted, since both are what firmware writes when it does not know.

## What still blocks link

```
cgem0: could not retrieve pclk.
cgem0: could not retrieve hclk.
cgem0: warning: attaching PHYs failed
```

The clock failures are **not** fatal and are expected — `cgem` deliberately tolerates them, and the MDC divider is a fixed `CGEM_NET_CFG_MDC_CLK_DIV_48` rather than derived from `pclk`.

The PHY is the real blocker. `mii_attach(..., MII_PHY_ANY, MII_OFFSET_ANY, 0)` scans every address and finds nothing, even though the tree describes one:

```
ethernet@100000/
    phy-mode = "rgmii-id"
    phy-handle -> ethernet-phy@1
    phy-reset-gpios
    phy-reset-duration
    ethernet-phy@1/          <- a direct child, not under an mdio node
```

`phy-reset-gpios` is the strong suspect: the PHY likely needs a reset pulse before it will answer on MDIO, and RP1's GPIO block has no driver yet. That is the next thing to establish, and it is not a guess I want to bake into a patch.

Everything underneath is proven in the same boot — PCIe, MSI-X, the RP1 interrupt controller, address translation and DMA all work, and the USB stack comes up on the same bus.

### DIFF
--- a/patches/0023-cgem-match-rp1-cadence-gem.patch
+++ b/patches/0023-cgem-match-rp1-cadence-gem.patch
@@ -1,0 +1,48 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: pkgdemon <jpm820@gmail.com>
+Date: Mon, 24 Aug 2026 00:15:00 -0500
+Subject: [PATCH] cgem: match RP1's Cadence GEM
+
+RP1, the southbridge on Raspberry Pi 5-family boards, presents a Cadence
+GEM behind PCIe. Nothing matched it, so the Pi 500+ had no ethernet:
+
+  rp10: <ethernet@100000> mem 0xc040100000-0xc040103fff irq 166
+      compat raspberrypi,rp1-gem (no driver attached)
+
+cgem(4) is already compiled in on arm64 -- GENERIC includes std.xilinx,
+which carries `device cgem` -- so this is only a question of matching.
+
+HWQUIRK_NONE rather than HWQUIRK_NEEDNULLQS, on the evidence of the
+vendor driver rather than by picking the safer-sounding one.  RP1's
+macb_config in Linux differs from zynqmp's in exactly two ways: it adds
+MACB_CAPS_CLK_HW_CHG and it omits MACB_CAPS_BD_RD_PREFETCH. Descriptor
+prefetch is the reason the Zynq parts need the null queue pointers
+programmed at all, so a part without it should not need them.
+
+Everything underneath this is already proven on the board: PCIe, MSI-X,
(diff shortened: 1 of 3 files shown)


pkgdemon:
MAC patch board-tested:

```
cgem0: <Cadence CGEM Gigabit Ethernet Interface> mem 0xc040100000-0xc040103fff irq 166 on rp10
cgem0: could not retrieve pclk.
cgem0: could not retrieve hclk.
cgem0: warning: attaching PHYs failed
cgem0: bpf attached
cgem0: Ethernet address: 88:a2:9e:49:6e:b3
```

The real address, from the device tree — matching what Linux uses for `eth0` on this board. The `no mac address found, assigning random` line is gone.

PHY blocker is now identified and filed as #104: the PHY sits behind a reset line on RP1's own GPIO controller (`phy-reset-gpios = <&gpio 32 GPIO_ACTIVE_LOW>`, 5 ms), which has no driver, so it never answers MDIO.

PR #123, opened 2026-08-28
mach: wire mach_port_get_attributes as a syscall trap (unblocks on-demand launch)

pkgdemon:
Wires `mach_port_get_attributes` as a direct syscall trap. **Prerequisite for
`nextbsd-userland#79`** (on-demand Mach-service launch), which cannot be fixed
without it.

## Why

launchd's demand scan asks the kernel which member ports of `demand_port_set`
have queued messages. libmach's `mach_port_get_attributes` is a stub returning
`KERN_RESOURCE_SHORTAGE`, because no such trap is exposed:

```
$ sysctl mach.syscall.mach_port_get_attributes
NOT PRESENT
```

So `mportset_callback()` can never identify a hot port, no job is kickstarted,
and the client blocks forever waiting for a reply nobody sends. Measured on
hardware: **210 seconds** with a demand message queued on
`com.apple.system.logger` and no spawn attempt, while `launchctl start` returned
instantly in the same window.

## What this actually adds

Very little — **the kernel implementation already exists and is complete.**
`ipc/mach_port.c`'s `mach_port_get_attributes()` handles
`MACH_PORT_RECEIVE_STATUS` and populates `mps_msgcount`, the single field
launchd reads. It was reachable only through MIG. This adds the direct trap
around existing, working code.

## Scope

**`MACH_PORT_RECEIVE_STATUS` only.** The other flavors are implemented but have
no in-tree caller; wiring them would add untested surface for no consumer.

**`mach_port_get_set_status` is deliberately NOT wired**, even though it is the
function launchd literally calls today. It is the only trap in the set returning
out-of-line memory, and libmach's `vm_deallocate` is a no-op (it is also called
on `malloc`'d `mig_allocate` buffers) — so a caller would leak a page per wakeup
in PID 1, forever, and fixing that drags an unrelated `mig_allocate` provenance
refactor into scope. launchd will instead enumerate its own `port_hash[]`, which
it already owns, and ask the kernel only for the message count. That also
repairs `job_keepalive`'s dead queued-message check (`core.c:6404`) for free.

## Conformance

Follows the existing pattern end to end:

- `PADL_/PADR_` args struct — bare fields silently corrupt arguments
- Task implicit via `current_task()->itk_space`, keeping this to 4 args and well
  inside the 6-argument ceiling
- Guarded wrapper with `mach_task_init_lazy` + `KERN_INVALID_ARGUMENT`, matching
  `mach_port_move_member`
- Sysctl uses the userland name; kernel symbol keeps its own
- All six wiring edits in `mach_syscall_wire.c`

**Register/deregister symmetry verified** — 15 each, exact mirror, new entry
appended last in register and prepended first in deregister.

## Also in this PR

**`.claude/skills/mach-syscall/SKILL.md`** — the conventions I had to reverse
engineer to write this, so the next person doesn't: the `patches/` vs
`src-overlay/` rule and the `config/NEXTBSD` trap that broke the modules build
in August; the full add-a-trap procedure; and the build/validation loop
including the fact that green CI is not evidence (the boot test is
`continue-on-error`, gated only on reaching `login:`).

It also records two things worth knowing that cost me time here: that
`mach_module.c`'s `osx_syscalls[]` is dead code for registration, and that
several trap handlers already exist in `mach_traps.c` unwired
(`_kernelrpc_mach_port_mod_refs_trap`, `_kernelrpc_mach_port_insert_member_trap`)
— the `mod_refs` one being a plausible cause of stale ports lingering in
`demand_port_set`.

**`patches/README.md`** — two stale `releng/15.0` references corrected to
`releng/15.1`, which is what `build.yml:15` actually pins, including a claim
that the patches were "verified to apply cleanly to releng/15.0".

## Testing

**Not yet booted.** CI is the first exercise; I will install the arm64 kernel on
a live VM and verify before this is merged.

The check is a clean single variable — on the current kernel:

```
$ sysctl mach.syscall.mach_port_get_attributes
NOT PRESENT
```

and after, it should return a slot number. Then the real test: stop syslogd,
send to `com.apple.system.logger`, and confirm launchd spawns it — the exact
experiment that failed for 210 s.

**Review focus:** the `copyout` bounds in the handler. `count` is validated
against `MACH_PORT_RECEIVE_STATUS_COUNT` before the call and reset to it after,
so the copyout length is fixed, but a second pair of eyes on that is worth having
in PID-1-adjacent code.

### DIFF
--- a/.claude/skills/mach-syscall/SKILL.md
+++ b/.claude/skills/mach-syscall/SKILL.md
@@ -1,0 +1,258 @@
+---
+name: mach-syscall
+description: Working in the NextBSD kernel repo — where a change belongs (patches/ vs src-overlay/), how to add or wire a Mach syscall trap, and how to build and validate a kernel change without a hot-reload loop. Use when touching src-overlay/sys/compat/mach, adding a trap, changing config/NEXTBSD, or wondering why a change compiled but nothing happened.
+---
+
+# Kernel changes and Mach syscall traps
+
+## Where a change belongs
+
+**`patches/` patches FreeBSD's own kernel. `src-overlay/` is NextBSD's own kernel code.**
+
+That is the whole rule, and it is about *intent*, not about whether the file is
+new — three patches (`0015`, `0018`, `0025`) add brand-new files, because they
+are FreeBSD-shaped drivers.
+
+| Change | Goes in |
+|---|---|
+| Edit an existing freebsd-src file | `patches/` + a line in `patches/series` |
+| New FreeBSD-shaped driver, wired via `sys/conf/files*` | `patches/` (wire it in the same patch) |
+| Anything under `compat/mach`, `dev/iokit`, `sys/sys/mach`, `sys/apple`, `kern/vfs_pivot.c` | `src-overlay/sys/…` + a line in the matching `src-overlay/conf/files.*` |
+| Option stock FreeBSD declares (`FUSEFS`, `UNIONFS`, …) | `config/NEXTBSD` |
+| Option only the overlay declares (`COMPAT_MACH`, `COMPAT_LINUX`, `LINPROCFS`, `LINSYSFS`) | a wiring step in `build.yml` — **never** `config/NEXTBSD` |
+
+### The trap in that last row
+
+`nextbsd-kernel-modules` reuses `patches/series` and `config/NEXTBSD` to build a
+KBI-matched kernel, and it **never lays in `src-overlay/`**. So an option the
+overlay declares makes that repo's `config(8)` die with `unknown option`. Naming
+them in the shared file is what broke both legs of the modules build on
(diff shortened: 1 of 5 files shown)


pkgdemon:
Closing. XNU has no `mach_port_get_attributes` trap — it is MIG-only on Apple, served from the task port — so this adds a divergence rather than restoring Apple's shape.

The kernel already ships the MIG server (`_Xmach_port_get_attributes` in `mach_port_server.c`), so the Apple-shaped fix is a MIG **client** in userland, and needs no kernel change at all. Feasibility of that is being verified now; if it holds, the whole fix lands in one `nextbsd-userland` PR.

The skill and the `releng/15.1` README correction are split out into a separate docs-only PR — they stand regardless of which approach wins.

PR #124, opened 2026-08-28
docs: skill for kernel changes and Mach syscall traps

pkgdemon:
Documentation only — no code, no behaviour change.

Adds `.claude/skills/mach-syscall/SKILL.md`, covering how to modify the Mach
module and how to patch the kernel itself. These conventions currently have to
be reverse-engineered from `build.yml` step comments, and getting them wrong
fails in ways that are quiet rather than loud.

## What it documents

**Where a change belongs.** `patches/` patches FreeBSD's own kernel;
`src-overlay/` is NextBSD's own kernel code. Including the sharp edge: an option
only the overlay declares must never go in `config/NEXTBSD`, because
`nextbsd-kernel-modules` reuses that file *without* laying in the overlay and
`config(8)` dies on it — which is what broke both legs on 2026-08-03.

**How to add or wire a Mach syscall trap.** Args struct, handler, files
fragment, `compat_shim` stub, the six edits in `mach_syscall_wire.c`, and the
register/deregister mirror with a copy-pasteable check for it.

**How to build and validate.** That there is no hot-reload loop — Mach is
compiled into the kernel, `NO_MODULES=yes`, no `mach.ko` exists — so every
change costs a full build and a reboot. How to pull an installable artifact from
a PR run, and how to install one on a live VM, including making a
`/boot/kernel.old` fallback **first**, since images ship without one and the
loader would otherwise have nothing to escape to.

## Three things recorded because each cost real time

- **`mach_module.c`'s `osx_syscalls[]` is dead code for registration.**
  `compat_shim.h` defines every `SYS_<name>` as `-1`, and
  `kern_syscall_helper_register` treats `NO_SYSCALL` as an end-of-array
  sentinel — so the loop exits on the first entry and registers nothing. Adding
  a trap there does nothing at all, silently.
- **Args structs must use `PADL_/PADR_`.** Bare fields silently corrupt
  arguments, because 4-byte fields do not line up with the kernel's 8-byte
  `register_t` slot writes.
- **Green CI is not evidence.** The boot test is `continue-on-error: true` and
  the only gate is grepping `boot.log` for a getty `login:` line. A run has
  shipped green with `FBSDGLUE-FAIL` and roughly twenty later markers never
  executed. Read the `boot-log` artifact and compare the marker census.

It also notes that several trap handlers already exist in `mach_traps.c` unwired
(`_kernelrpc_mach_port_mod_refs_trap`, `_kernelrpc_mach_port_insert_member_trap`),
and that implementations exist in `ipc/mach_port.c` reachable only through MIG —
so check what is already there before writing a new one. The `mod_refs` one is a
plausible cause of stale ports lingering in `demand_port_set`.

## Also

Two stale `releng/15.0` references in `patches/README.md` corrected to
`releng/15.1`, which is what `build.yml:15` actually pins — including a claim
that the patches were "verified to apply cleanly to `releng/15.0`".

## Note on scope

This was split out of a since-closed PR that also wired
`mach_port_get_attributes` as a syscall trap. That trap was dropped because XNU
has no such trap — `mach_port_get_attributes` is MIG-only on Apple — and the
Apple-shaped fix belongs in userland instead. The skill's guidance on *how* to
add a trap stands on its own regardless, and is written generically from the
existing `mach_wait_quiet` and `mach_port_move_member` examples.

Formatted to match the existing `rpi-bringup` skill.

### DIFF
--- a/.claude/skills/mach-syscall/SKILL.md
+++ b/.claude/skills/mach-syscall/SKILL.md
@@ -1,0 +1,258 @@
+---
+name: mach-syscall
+description: Working in the NextBSD kernel repo — where a change belongs (patches/ vs src-overlay/), how to add or wire a Mach syscall trap, and how to build and validate a kernel change without a hot-reload loop. Use when touching src-overlay/sys/compat/mach, adding a trap, changing config/NEXTBSD, or wondering why a change compiled but nothing happened.
+---
+
+# Kernel changes and Mach syscall traps
+
+## Where a change belongs
+
+**`patches/` patches FreeBSD's own kernel. `src-overlay/` is NextBSD's own kernel code.**
+
+That is the whole rule, and it is about *intent*, not about whether the file is
+new — three patches (`0015`, `0018`, `0025`) add brand-new files, because they
+are FreeBSD-shaped drivers.
+
+| Change | Goes in |
+|---|---|
+| Edit an existing freebsd-src file | `patches/` + a line in `patches/series` |
+| New FreeBSD-shaped driver, wired via `sys/conf/files*` | `patches/` (wire it in the same patch) |
+| Anything under `compat/mach`, `dev/iokit`, `sys/sys/mach`, `sys/apple`, `kern/vfs_pivot.c` | `src-overlay/sys/…` + a line in the matching `src-overlay/conf/files.*` |
+| Option stock FreeBSD declares (`FUSEFS`, `UNIONFS`, …) | `config/NEXTBSD` |
+| Option only the overlay declares (`COMPAT_MACH`, `COMPAT_LINUX`, `LINPROCFS`, `LINSYSFS`) | a wiring step in `build.yml` — **never** `config/NEXTBSD` |
+
+### The trap in that last row
+
+`nextbsd-kernel-modules` reuses `patches/series` and `config/NEXTBSD` to build a
+KBI-matched kernel, and it **never lays in `src-overlay/`**. So an option the
+overlay declares makes that repo's `config(8)` die with `unknown option`. Naming
+them in the shared file is what broke both legs of the modules build on
+2026-08-03. `config/NEXTBSD:114-122` says so; believe it.
+
+Mechanically: `patches/` is `git apply`'d first (`build.yml:37-43`), then
+`src-overlay/sys/` is `cp -R`'d over `/usr/src/sys/` (`build.yml:111`), then five
+wiring steps `cat` fragments from `src-overlay/conf/` onto `sys/conf/files*`.
+A patch can therefore never depend on an overlay file; the overlay may assume
+patched base sources, and does.
+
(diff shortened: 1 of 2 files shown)


pkgdemon:
Holding this for a revision rather than merging as-is.

The repo-conventions half is correct and worth keeping — `patches/` vs `src-overlay/`, the `series` ordering, provenance checks, "green CI is not evidence", the build/test recipes.

But the **"Adding a Mach syscall trap"** section teaches trap-wiring as *the* pattern, and that is now the wrong default. Measured since this PR was opened (nextbsd-kernel#125):

- Apple's shape is **MIG as the floor, traps as an optional accelerator** that falls back to MIG on `MACH_SEND_INVALID_DEST`. Routines returning variable-length arrays have no trap at all.
- Traps cost **one scarce syscall slot each**. `patches/0001` already widened FreeBSD's `lkmnosys` band from 10 slots to 58 because "slots (210-219) are exhausted by the Mach trap set". ~156 kernel RPC routines would need ~156 slots. **MIG costs zero** — everything rides the single already-wired `mach_msg_trap`.

Suggested change: keep the trap procedure, but put a decision gate in front of it — *does Apple use MIG here, and does a MIG path exist?* — with the trap documented as an accelerator layered on a working MIG client, never as the first resort. That would have steered #123 correctly.

pkgdemon:
Closing rather than merging. Roughly half of this is correct and worth keeping; the other half teaches a default the MIG work has since disproved, and merging it would enshrine that.

**The problem:** the "Adding a Mach syscall trap" section presents trap-wiring as the way to reach the kernel. Measured since this PR was opened (#125):

- `patches/0001` widened FreeBSD's `lkmnosys` band from **10 slots to 58** — in its own words because "lkmnosys slots (210-219) are exhausted by the Mach trap set".
- There are ~156 kernel RPC routines across the seven Mach subsystems. Traps cost **one slot each**. That does not fit and never could.
- **MIG costs zero slots** — everything rides the single already-wired `mach_msg_trap`.

So the 10→58 widening was a symptom of an approach with no ceiling, not a fix. This section would have steered #123 the same wrong way a second time.

**What was right here and is preserved in #129:** where a change belongs (`patches/` vs `src-overlay/`), provenance checks, the build/test recipes, "green CI is not evidence", and publishing. Plus the trap procedure itself — kept, but behind a decision gate (*does Apple use MIG here, and does a MIG path exist?*), with traps documented as an accelerator over a working MIG client rather than the first resort. 9 of our 12 traps do match a real Apple trap and should stay.

**Why a rewrite rather than a fixup:** the ground moved. `mach_port` now generates from its `.defs` at build time (#128), which brings in the `.msgids` ABI contract, the generation and verification scripts, the fact that the frozen servers are hand-adapted in ways pure MIG output cannot reproduce, and the trap that passing the ABI gate does **not** mean the file will compile. A doc written before any of that existed is better replaced than patched.

Tracked in #129. The branch content is recoverable from this PR if it is easier to start from.

Issue #131, opened 2026-08-29
Generated mach_port server passes the raw port instead of an ipc_space_t — page fault in ipc_right_lookup


pkgdemon:
## Userland can panic the kernel by destroying a port set with kevents attached

Reproduced on **both arches** in nextbsd-userland CI:

```
Fatal trap 12: page fault while in kernel mode
cpuid = 0; apic id = 00
fault virtual address	= 0x488
fault code		= supervisor read data, page not present
instruction pointer	= 0x20:0xffffffff80b974a2
current process		= 42 (test_evfilt_machpor)
rsi: 0000000000000000  rbx: 0000000000000000  r14: 0000000000000000
...
#12 0xffffffff80f8ed20 at sys_mach_msg_trap+0x30
#13 0xffffffff810db606 at amd64_syscall+0x126
#14 0xffffffff810b17cb at fast_syscall_common+0xf8

Uptime: 1s
```

The low fault address (`0x488`) and the null registers say this is a null-ish dereference at a struct offset — a port set read after it has been freed, or freed under a filter that still references it.

## Reproducer

`src/mach_kmod/tests/test_evfilt_machport_concurrent.c` — the existing `EVFILT-MACHPORT-CONCURRENT` stress test. Line 146 calls `mach_port_mod_refs()` on a **port set** while worker threads concurrently attach, detach and destroy kevents against it and churn `mach_port_move_member`.

## Why it has never fired before

`mach_port_mod_refs()` in libmach is a stub that returns `KERN_SUCCESS` without doing anything (nextbsd-userland#83). With reference counting silently disabled, the port set was **never actually destroyed**, so the race had nothing to race against. The test has been passing for that reason, not because the path is sound.

Wiring `mach_port_mod_refs` to the real MIG client — one line — panics the kernel within a second of the test starting.

## Status

`mach_port_mod_refs` is **deliberately left as a no-op** in nextbsd-userland#83 for exactly this reason, and split out from the six routines that were converted. It becomes a real RPC once this is fixed.

That is a workaround in userland for a kernel defect, and it should not stand: **userland must never be able to panic the kernel**, regardless of what it asks for. Same class as #391 in the nextbsd repo.

## Worth noting

nextbsd-kernel#125 predicted this specific failure before it happened:

> Making `mach_port_mod_refs` real is a behavioural change, not a bug fix. It currently no-ops. Once it actually drops references, any latent over-release or missing-retain in the tree stops being harmless and starts destroying ports.

[...]

Issue #127, opened 2026-08-28
Reconstruct the remaining six MIG subsystems (task, vm_map, host_priv, mach_host, mach_vm, clock)


pkgdemon:
Part of #125. **Phase 7** — after #126 proves the build wiring.

Six subsystems remain frozen. Same disease, same cure, one PR and one dispatch-table gate each.

- [ ] `task` — base 3400, 42 slots — needs `mach_zone_name_array_t`
- [ ] `vm_map` — base 3800, 31 slots — needs `vm_map_t`, macro-wrapped routine names
- [ ] `host_priv` — base 400, 26 slots — needs `vm_map_t`
- [ ] `mach_host` — base 200, 25 slots — needs `clock_id_t`
- [ ] `mach_vm` — base 4800, 20 slots — needs `vm_map_t`, macro-wrapped routine names
- [ ] `clock` — base 1000, 3 slots — needs `clock_serv_t`

## What was measured

The `mach_port` pipeline was run against all six. All six currently fail, but **every failure is shallow and nothing is absent from Apple's open-source `.defs`**:

1. **Missing type vocabulary** — the same gap that required writing `mach_debug_types.defs` from scratch. Our `mach_types.defs` is a subset of what these subsystems reference.
2. **Macro-wrapped routine names** — unlike `mach_port`, these declare routines as e.g.

   ```
   #if !KERNEL && !LIBSYSCALL_INTERFACE && !KERNEL_VM_TEST
   skip;
   #else
   routine PREFIX(KERNEL_SERVER_SUFFIX(vm_allocate))(
   ```

   so names only resolve *after* cpp. Name extraction must run on preprocessed output, not raw text. `mach_port` happened to be the one subsystem without this pattern, which is why it went first and went smoothly.

**These are six real units of work, not a rounding error on the first one.** Sequenced last deliberately: each is mechanical once #126 has proved the build wiring, and landing 27,455 lines of deletion across six unproven reconstructions in one PR would be neither reviewable nor bisectable.

Full plan: https://pkgdemon.github.io/nextbsd-mach-mig-restoration-plan.html

pkgdemon:
Scope estimate raised, based on actually converting `mach_port` (#126).

This ticket described the remaining six as needing type-vocabulary extensions and macro-aware name resolution. Both are still true, but they are not the whole job.

The 2015 servers are **hand-adapted**, and pure MIG output does not reproduce that:

[...]

Issue #125, opened 2026-08-28
Restore MIG as the kernel RPC substrate — 7 subsystems, 27,455 lines of unregenerable generated C


pkgdemon:
## Summary

Seven kernel MIG servers were generated in 2015 by a MIG we no longer have, and checked in as source **with no `.defs` beside them**. There is no source of truth for the kernel's RPC wire protocol, so neither side can be regenerated and the two ends cannot be guaranteed to agree — which is the one thing MIG exists to provide.

Userland then papered over the resulting gap: roughly **18 `libmach` entry points return `KERN_SUCCESS` having done nothing**, including both calls that on-demand Mach-service launch depends on (nextbsd-userland#79).

| Generated server (checked in) | Lines | Base | Slots | Reconstructs today? |
|---|---:|---:|---:|---|
| `mach_port_server.c` | 5,720 | 3200 | 36 | **yes — verified** |
| `task_server.c` | 6,376 | 3400 | 42 | needs `mach_zone_name_array_t` |
| `vm_map_server.c` | 4,444 | 3800 | 31 | needs `vm_map_t` + macro-wrapped names |
| `host_priv_server.c` | 4,022 | 400 | 26 | needs `vm_map_t` |
| `mach_host_server.c` | 3,199 | 200 | 25 | needs `clock_id_t` |
| `mach_vm_server.c` | 3,028 | 4800 | 20 | needs `vm_map_t` + macro-wrapped names |
| `clock_server.c` | 666 | 1000 | 3 | needs `clock_serv_t` |
| **total** | **27,455** | | **183** | |

## The kernel MIG path already works — measured, not assumed

Two hand-built messages sent to `mach_task_self()` on `gershwin-on-nextbsd`, compiled natively:

```
probe 1  mach_port_get_attributes (3217)
  reply 3317  RetCode=0  outCnt=10
  mps_pset=0 seqno=0 mscount=0 qlimit=5 MSGCOUNT=0    <- qlimit=5 is MACH_PORT_QLIMIT_DEFAULT,
                                                          i.e. real kernel state, not a zeroed buffer

probe 2  mach_port_get_set_status (3214)              <- the exact call libmach fakes
  reply 3314  size=56  complex=yes  desc_count=1
  membersCnt=1  ool.size=4  ool.address=0x1000
  member[0] = 0x12                                    <- the port we had just inserted
```

[...]

PR #134, opened 2026-08-29
tools: commit the .defs reconstructor used to recover mach_port.defs

pkgdemon:
`mach_port.defs` was recovered from the frozen 2015 server by a script that merged its **dispatch table** (for order, which *is* the ABI) with **Apple's declarations** (for signatures, which generated C cannot carry). That script was never committed — so #127, six more subsystems needing exactly the same treatment, would have had to rebuild it from scratch.

## To be clear about what is generated

A `.defs` is **hand-authored source**, same as Apple's `osfmk/mach/*.defs`. The build generates C *from* `.defs`; nothing generates `.defs`. This tool is one-shot archaeology for recovering a `.defs` that never existed — not part of the build, and not needed to maintain one afterwards.

## Verified before committing

- re-running it against the frozen `mach_port_server.c` recovered from git history reproduces the committed `mach_port.msgids` **identically**
- running it on `task_server.c` (still frozen) yields 42 slots at base 3400 and **warns** about six routines it cannot match in Apple's `task.defs`, rather than silently emitting a wrong file

That warning behaviour is the point. Those six are the macro-wrapped names (`PREFIX(KERNEL_SERVER_SUFFIX(...))`) already flagged on #127, and a reconstructor that skipped them quietly would produce a `.defs` that **passes the msgid gate**, renumbers nothing, and omits real routines.

## The docstring carries the two traps that cost the most

1. `KERNEL_SERVER` must be **defined** at generation time even though we collapse its `*_from_user` names, because it also activates the `intran`/`destructor` translations. Without it the dispatch passes the raw request port where an `ipc_space_t` is required and page-faults in `ipc_right_lookup()` — and the ABI gate cannot see it, because the wire format is unchanged (#131, #133).
2. The frozen servers carry a hand-adapted include prologue that pure MIG output does not reproduce; `ci/gen-mig-servers.sh` reapplies it.

Refs #125, #127.

### DIFF
--- a/tools/reconstruct-defs.py
+++ b/tools/reconstruct-defs.py
@@ -1,0 +1,159 @@
+#!/usr/bin/env python3
+"""
+reconstruct-defs.py -- recover a MIG .defs from a frozen generated server.
+
+ONE-SHOT ARCHAEOLOGY TOOL. Not part of the build, and not needed to maintain
+a .defs once it exists -- a .defs is hand-authored source from then on, the
+same as Apple's own osfmk/mach/*.defs.
+
+It exists because sys/compat/mach carries seven MIG servers that were
+generated in 2015 by a MIG we no longer have, and checked in as source with
+no .defs beside them (nextbsd-kernel#125). This reconstructs the missing
+input so both ends can be generated from one description again.
+
+How it works, and why it is split this way:
+
+  ORDER comes from the frozen server's dispatch table. msgh_id is subsystem
+  base + routine index, so the order IS the ABI. NextBSD's ordering is its
+  own -- it differs from XNU's in 34 of 36 positions for mach_port -- so
+  Apple's .defs cannot be substituted wholesale.
+
+  SIGNATURES come from Apple's .defs for the same subsystem, because they
+  carry the type qualifiers (intran, CountInOut, inout, dealloc) that cannot
+  be recovered from generated C. Verify them against the frozen server's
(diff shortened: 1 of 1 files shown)


pkgdemon:
Closing. Two reasons, and the second is the serious one.

**1. Apple does not do this.** `osfmk/mach/mach_port.defs` is hand-maintained source, ~19KB of it. There is no reconstruction tool in XNU because Apple never lost the source. We needed the archaeology exactly once, because the 2015 servers were committed without their `.defs`. That job is done.

**2. Committing it is actively hazardous.** I checked what re-running it would produce against what is committed:

```
committed mach_port.defs : 368 lines
tool output              : 271 lines
KERNEL_SERVER header     : ABSENT from tool output
```

The committed `.defs` carries hand-written documentation explaining that `KERNEL_SERVER` must be defined at generation time — the thing whose absence page-faulted the kernel in `ipc_right_lookup()` (#131/#133). The tool does not emit it. So anyone re-running this to "refresh" the `.defs` would silently drop the warning and could easily reintroduce the panic.

A tool that can quietly destroy hand-maintained source is worse than no tool.

**Nothing is lost.** The method and both traps are already written up as prose on #127 — the order/signature split, the `KERNEL_SERVER` double duty, the macro-wrapped names, and the extra `grep -c 'In0P->Head.msgh_request_port'` check. That is the durable form for knowledge like this; a script is not.

`mach_port.defs` stands on its own as hand-authored source from here, the same as Apple's.

PR #135, opened 2026-08-29
mach_msg: report the trailer the MIG server stub actually sees (#91)

pkgdemon:
Diagnostic only, no behaviour change.

#91 is not a lost wakeup. Measured chain: `notify_post` -> `NOTIFY_STATUS_FAILED` -> real code 11 `SERVER_CHECKIN_FAILED` -> raw `kstatus=0xfffffecb` = **-309 `MIG_TRAILER_ERROR`**, against a port already proven good (`bootstrap_look_up` returns `kr=0, port=0x13`).

`notify_ipc.defs` declares `ServerAuditToken` on the checkin routines, so the check runs in **notifyd's server stub** against the request message. syslogd then blocks in `ipc_mqueue_receive` because its checkin genuinely failed.

migcom emits two ways to reach that error:
```
TrailerP->msgh_trailer_type != MACH_MSG_TRAILER_FORMAT_0
trailer_size < sizeof(audit_token_t)
```
with `TrailerP = In0P + round_msg(In0P->Head.msgh_size)`.

Three hypotheses already died on inspection (uninitialised `au_tok`; `LEGACY_HEADER_SIZE_DELTA` offset skew; `dsc_adjust` moving the trailer — it self-cancels). This prints what the stub will actually read, so the next boot answers it instead of a fourth guess.

Rate-limited to 40 lines. Refs #91.

### DIFF
--- a/src-overlay/sys/compat/mach/ipc/mach_msg.c
+++ b/src-overlay/sys/compat/mach/ipc/mach_msg.c
@@ -487,6 +487,51 @@ mach_msg_receive(
 		FREE_SCATTER_LIST(slist, slist_size, slist_rt);
 		return mr;
 	}
+	/*
+	 * #91 diagnostic. notifyd's MIG server stub rejects the checkin request
+	 * with MIG_TRAILER_ERROR (-309, seen as kstatus=0xfffffecb in
+	 * libnotify), which it returns from exactly two checks:
+	 *
+	 *	TrailerP->msgh_trailer_type != MACH_MSG_TRAILER_FORMAT_0
+	 *	trailer_size < sizeof(audit_token_t)
+	 *
+	 * where the stub locates the trailer itself as
+	 *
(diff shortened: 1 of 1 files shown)


pkgdemon:
This diagnostic did its job: it identified #91's root cause in a single boot after four hypotheses derived from reading the source had all been wrong.

```
MIGTRAILER: id=1023 size=32 opt=0x106 type=0 tsize=8 moved=0 complex=0
```

`opt=0x106` = `RCV_MSG|RCV_LARGE|RCV_TIMEOUT` with an empty trailer field, so the kernel never grew the trailer and it stayed at the 8-byte minimum; the MIG server stub then failed `8 < sizeof(audit_token_t)`. Fixed userland-side in nextbsd-userland#99.

Keeping the check rather than reverting it: it is rate-limited to 40 lines, only fires when a delivered trailer is malformed, and would have caught this class instantly. It is a cheap standing assertion on the MIG trailer contract.

Issue #136, opened 2026-08-30
ipc_port_release_send() asserts IP_VALID, but modern migcom emits it unguarded — will panic on mach_vm conversion


pkgdemon:
Found while reconstructing `mach_vm.defs` for #127. Not a `.defs` problem — a kernel one, and currently latent.

## The mismatch

The frozen 2015 stubs guard every right-release:

```c
if (IP_VALID((ipc_port_t)In0P->parent_handle.name))
        ipc_port_release_send((ipc_port_t)In0P->parent_handle.name);
```

The migcom we use now emits it **unguarded**, because modern XNU moved the validity check inside the function:

```c
ipc_port_release_send((ipc_port_t)In0P->parent_handle.name);
```

But our `ipc_port_release_send()` (`sys/compat/mach/ipc/ipc_port.c:1217`) still has the *old* contract — it opens with `assert(IP_VALID(port))` and then `ip_lock(port)`, with no early return for `MACH_PORT_NULL`.

## Why it has not bitten yet

`mach_port` was the first subsystem converted to generated output (#128) and has no `in` port arguments released this way, so nothing exercised the path.

`mach_vm` does. `_mach_make_memory_entry` (msgid 4817) takes `parent_handle` as an `in` port, and **passing `MACH_PORT_NULL` there is legitimate and common** — it means "no parent entry". Under the current stubs that is fine; under generated stubs it goes straight into `assert(IP_VALID(port))`.

So converting mach_vm to generated output — which is the goal of #127 — turns a normal call into a kernel assert.

## Fix

Bring `ipc_port_release_send()` in line with the contract its callers now assume: return early on `!IP_VALID(port)` instead of asserting. That matches modern XNU and makes both the frozen and generated stubs correct.

Worth auditing the sibling functions (`ipc_port_release_receive`, `ipc_port_release_sonce`) for the same assumption while in there.

## Pattern

This is the same shape as `MACH_PORT_TEMPOWNER` (#104) and the aslmanager trigger (#80): a caller-side guard disappears, and the callee was relying on it without saying so. Each one stays invisible until something makes the path real.

Blocks the mach_vm half of #127.

Issue #93, opened 2026-08-24
rpi500: /chosen/bootargs is not reaching the kernel — boot flags and tunables from cmdline.txt are inert


pkgdemon:
Boot flags passed via `cmdline.txt` have no effect on this board, which quietly removes an instrument.

## What was observed

`tryboot.txt` names a command line file:

```
kernel=kernel8.img
device_tree=bcm2712-rpi-500.dtb
cmdline=nextbsd-cmdline.txt
arm_64bit=1
enable_uart=1
```

and `nextbsd-cmdline.txt` contains exactly:

```
FreeBSD: -v
```

The resulting boot is **not verbose** — no `map[10]:` BAR decodes, no `domain=`, and the whole serial log is 35 KB where a verbose arm64 boot runs several times that.

## Why it matters

`RB_VERBOSE` is how you see PCI BAR assignment, capability decoding, and interrupt routing. On a board where [everything fails as silence](https://pkgdemon.github.io/nextbsd-rpi500-native-boot-plan.html), losing it costs real diagnostic ground — the RP1 MSI-X failure in #81 had to be diagnosed by reading the endpoint from Linux instead, which only worked because this board dual-boots.

It is not only `-v`. `boot_parse_arg()` sets **any** `name=value` token as a kernel environment variable, so this same path is how a loader-less board would set tunables at boot without a rebuild. All of it is currently inert.

## Where it goes wrong

The prefix is right — `sys/arm64/arm64/machdep_boot.c` defines `CMDLINE_GUARD "FreeBSD:"` and `cmdline_set_env()` matches it case-insensitively, then does `boothowto |= boot_parse_cmdline(cmdline)`.

The gate is upstream of that:

```c
void
parse_fdt_bootargs(void)
{
	if (loader_envp == NULL && fdt_get_chosen_bootargs(linux_command_line,
	    LBABI_MAX_COMMAND_LINE) == 0) {
		init_static_kenv(static_kenv, sizeof(static_kenv));
		cmdline_set_env(linux_command_line, CMDLINE_GUARD);
	}
}
```

`loader_envp` is NULL on route (a) — there is no loader — so the live term is `fdt_get_chosen_bootargs()`, which is `OF_getprop(chosen, "bootargs", ...)`. The most likely reading is that **the firmware is not writing `/chosen/bootargs` into the DTB it hands over** under `tryboot`.

Worth ruling out before assuming that:

[...]

Issue #137, opened 2026-08-30
Two MIG wire sizes are stale on LP64: vm_info_object_t (84 vs 88) and page_address_array_t (4 vs 8)


pkgdemon:
Found while reconstructing `vm_map.defs` for #127. Reproduced as-frozen deliberately; fixing them changes the wire format, so it needs its own decision.

## The mismatch

Two types have a wire size that disagrees with our C header, both because `vm_offset_t` is 64-bit here and the MIG numbers predate that:

| type | wire | our C header | used by |
|---|---|---|---|
| `vm_info_object_t` | 84 bytes | 88 bytes | `mach_vm_region_info` |
| `page_address_array_t` element | 4 bytes | 8 bytes | `vm_mapped_pages_info` |

Both are inherited verbatim from Apple's `vm_map.defs`, where they were correct for a 32-bit `vm_offset_t`.

## Why they were reproduced rather than fixed

The reconstruction's whole premise is that the regenerated server must be ABI-identical to the frozen one. Correcting these sizes would change the wire format of two routines, which is exactly the kind of silent divergence #125 exists to eliminate. So they are reproduced as-is and documented in the `.defs`.

## What is actually wrong

A client marshals 88 bytes per `vm_info_object_t`; the server unmarshals 84. On an array that misaligns every element after the first. Same shape for `page_address_array_t` at 4-vs-8.

Neither routine appears to be exercised today — which is why this has gone unnoticed — but they are live dispatch slots, not tombstones.

## Options

1. Correct the `.defs` to match the 64-bit C types and accept a wire-format change on both routines. Correct, but a real ABI break.
2. Correct the C headers to match the wire. Wrong on LP64 — the fields genuinely are 64-bit.
3. Leave frozen, document, and gate: add both routines to a known-broken list so nothing starts depending on them.

Option 1 is probably right given nothing uses them yet, but it should be a deliberate call, not a side effect of #127.

Refs #127, #125.

PR #138, opened 2026-08-30
mach: reconstruct the six remaining MIG subsystems from the frozen servers (#127)

pkgdemon:
Closes the reconstruction half of #127, under #125 ("Restore MIG as the kernel RPC substrate").

**21,735 lines of generated C were committed as source with no `.defs` beside them.** This adds the missing descriptions plus frozen msgid contracts, so all six can be generated the way `mach_port` already is.

| subsystem | base–end | slots | real | tombstones |
|---|---|---|---|---|
| `mach_host` | 200–225 | 25 | 17 | 8 |
| `host_priv` | 400–426 | 26 | 22 | 4 |
| `clock` | 1000–1003 | 3 | 3 | 0 |
| `task` | 3400–3442 | 42 | 36 | 6 |
| `vm_map` | 3800–3831 | 31 | 25 | 6 |
| `mach_vm` | 4800–4820 | 20 | 17 | 3 |

Plus `clock_types.defs` (userland has no copy to `-I` at) and nine added types in `mach_debug_types.defs`.

## Verification

Each subsystem was regenerated with migcom built per `ci/gen-mig-servers.sh` (`-D__MACH30__ -DNEXTBSD_KERNEL_SERVER=1 -DKERNEL_SERVER=1`) and diffed against the committed server: **routine tables identical** slot-for-slot including every `skip;`, **every `__Request__`/`__Reply__` struct identical** field-for-field, and `verify-mig-abi.sh` passing against both the generated *and* the frozen server.

The residual textual diff is entirely migcom-version drift, and that was **proven rather than asserted** — by regenerating `mach_port` from its already-merged `.defs`, diffing against its pre-#128 frozen version to establish a baseline of differences the accepted conversion already produces, then checking every diff line-shape against that baseline. `verify-mig-abi.sh` was also confirmed to *fail* on a deliberately corrupted contract, so the check is live rather than vacuous.

## Why reconstruct rather than import Apple's .defs

**Using current XNU would have silently corrupted the kernel.** It wraps `vm_allocate_cpm` (406) and `host_get_clock_control` (408) in `#if KERNEL_SERVER skip;`. Those sit *mid-table*, so substituting Apple's file renumbers every routine above them — and the failure mode is the kernel dispatching the **wrong function**, not an error. Same trap at `mach_host` 214/223/224. Both `.defs` headers and `.msgids` record this.

## Findings worth acting on separately

- **kernel#136** — modern migcom emits `ipc_port_release_send()` unguarded, but ours still `assert(IP_VALID(port))`s. `_mach_make_memory_entry` legitimately passes `MACH_PORT_NULL`, so converting mach_vm to generated output would turn a normal call into a kernel assert. **Blocks the mach_vm conversion.**
- **kernel#137** — two stale wire sizes inherited from Apple (`vm_info_object_t` 84 vs 88, `page_address_array_t` 4 vs 8) because `vm_offset_t` is 64-bit here. Reproduced as-frozen deliberately.
- **For #93** — the servers those fabricating userland entry points wait on barely exist: `host_priv` is 3-of-22 real, `mach_host` is **1-of-17**. `host_info`, `host_page_size`, `host_kernel_version` all return `KERN_NOT_SUPPORTED`, and under `MACH_DEBUG && INVARIANTS` that **panics**.
- `vm_map` 3812 (`vm_map`) and 3826 (`vm_map_64`) are live in Apple and have never existed here — this kernel hands out named memory entries but has no MIG call to map one into an address space. A missing feature, not a numbering error; the holes stay.

## Scope

**Description only — no conversion.** The committed `*_server.c` files are untouched and `ci/gen-mig-servers.sh` is unchanged; wiring each subsystem up is a separate change, as it was for `mach_port` (#128). mach_vm's conversion additionally needs #136 first.

## One caveat on this branch

Three agents produced these independently and each verified its own work; I then cherry-picked all three and hand-resolved one conflict in `mach_debug_types.defs`, where they disagreed on `mach_zone_name_array_t` (inline vs out-of-line). I kept the out-of-line form since both `task_server.c:996` and `mach_host_server.c:587` declare `mach_msg_ool_descriptor_t`. **That merge is being re-verified across all six subsystems plus `mach_port` as a control**; I will post the result here. If task and mach_host turn out to need different forms, the type must be split rather than merged.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -176,9 +176,21 @@ jobs:
       - name: Verify MIG ABI against the frozen contract
         run: |
           set -eu
+          # mach_port is GENERATED, so this checks the generator's output.
           "$GITHUB_WORKSPACE/ci/verify-mig-abi.sh" \
             /usr/src/sys/compat/mach/mach_port_server.c \
             "$GITHUB_WORKSPACE/src-overlay/sys/sys/mach/mach_port.msgids"
+
+          # The other six are still the FROZEN committed servers (#127 added
+          # their .defs/.msgids; the conversion is separate). Checking them
+          # here pins the contract against the servers actually being built,
+          # so a bad `skip;` or a renumbering cannot land unnoticed. Without
+          # this, nothing in CI touches these six at all.
+          for sub in task vm_map host_priv mach_host mach_vm clock; do
+            "$GITHUB_WORKSPACE/ci/verify-mig-abi.sh" \
+              "/usr/src/sys/compat/mach/${sub}_server.c" \
+              "$GITHUB_WORKSPACE/src-overlay/sys/sys/mach/${sub}.msgids"
+          done
 
       - name: Wire in-kernel IOKit catalogue (K2, #215)
         run: |
--- a/src-overlay/sys/sys/mach/clock.defs
+++ b/src-overlay/sys/sys/mach/clock.defs
@@ -1,0 +1,101 @@
+/*
+ * clock.defs -- MIG interface for the clock subsystem.
+ *
+ * Reconstructed from src-overlay/sys/compat/mach/clock_server.c, generated in
+ * 2015 by a MIG we no longer have and checked in as source with no .defs
+ * beside it. See mach_port.defs for the full rationale; the short version is
+ * that a subsystem with no description has no guarantee its kernel server and
+ * its libmach client agree, which is the one thing MIG exists to provide.
+ * nextbsd-kernel#127, under #125.
+ *
+ * ROUTINE ORDER IS LOAD-BEARING. msgh_id is subsystem base + index, so the
+ * order below reproduces the frozen dispatch table: 1000, 1001, 1002. There
+ * are no `skip;` tombstones in this subsystem -- the three routines are
+ * contiguous and the descriptor's end is 1003, exactly base + 3.
+ *
+ * Unlike mach_port, this subsystem's order is Apple's order. NextBSD took
+ * clock.defs unchanged, and the reconstruction is a routine-for-routine match
+ * with Apple's mach/clock.defs.
+ *
+ * WHAT IS NOT HERE: Apple's clock subsystem has a sibling, clock_priv (base
+ * 1200: clock_set_time, clock_set_attributes), and clock_reply (base 3625:
+ * clock_alarm_reply). Neither was generated into this tree, so neither is
+ * reconstructed. clock_alarm can therefore arm an alarm that only a
+ * hand-written sender can deliver -- see the note on clock_reply_t in
+ * clock_types.defs.
+ *
+ * Signatures were cross-checked against the frozen __Request__/__Reply__
+ * structs and the frozen LINTLIBRARY prototypes: 3/3 agree.
+ */
+
+subsystem
+#if	NEXTBSD_KERNEL_SERVER
+	KernelServer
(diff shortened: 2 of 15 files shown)


pkgdemon:
## Re-verification found a regression I introduced in the merge — fixed in `44baee8`

Posting the result I promised, including the part where my resolution was wrong.

### My conflict resolution was backwards

I kept the out-of-line (`^`) form of `mach_zone_name_array_t`, reasoning that both `task_server.c:996` and `mach_host_server.c:587` declare `mach_msg_ool_descriptor_t`. **That reasoning is invalid — both forms emit a `MACH_MSG_OOL_DESCRIPTOR`, so the descriptor cannot distinguish them.**

The actual discriminator is the count zero-init. An unbounded `array[]` sets `itMigInLine`; migcom emits `WriteInitializeCount` (`server.c:1258`) only for inline types. `^array[]` runs `itPtrDecl`, clears the flag, and the init disappears:

```
_Xtask_zone_info    frozen: namesCnt=0, infoCnt=0   with ^: infoCnt only
_Xmach_zone_info    frozen: namesCnt=0, infoCnt=0   with ^: neither
```

Consequence: the kernel routine receives `&OutP->namesCnt` **uninitialised**, and the reply's OOL size becomes garbage × 80. The task/vm_map agent's original non-`^` form was correct.

Two more were wrong the same way, from the original commits rather than my merge: `mach_zone_info_array_t` and `hash_info_bucket_array_t`. `vm_info_object_array_t` and `page_address_array_t` are correctly `^` — confirmed by flipping them as a control, which produces three spurious `Cnt = 0` lines in vm_map. The corrected set matches Apple's own spelling.

The `^` semantics are now documented in the file so nobody "tidies" them later.

### Verification result (merged tree, before the fix)

| subsystem | routine table | structs | verify-mig-abi |
|---|---|---|---|
| task | identical | 36/36 + 36/36, **1 real field diff** | PASS |
| vm_map | identical | 25/25 + 25/25 clean | PASS |
| host_priv | identical | 22/22 + 22/22 clean | PASS |
| mach_host | identical | 17/17 + 17/17 clean | PASS |
| mach_vm | identical | 17/17 + 17/17 clean | PASS |
| clock | identical | 3/3 + 3/3 clean | PASS |
| **mach_port (control)** | — | — | **byte-identical** before/after my type merge |

mach_port being unchanged confirms the shared-type edit didn't disturb the already-converted subsystem.

### Also fixed

- `task_swap_exception_ports`' reply field is `old_handlerss` (doubled s) in `task_server.h:401`; `task.defs` had Apple's current spelling. ABI-neutral but not faithful. `task_get_exception_ports` keeps the single s.
- The wire-size rationale for #137 was lost in my cherry-pick. Restored.

### CI gap closed

`gen-mig-servers.sh` still ends at `gen_one mach_port`, so **nothing in CI touched these six** — which is precisely why the `^` regression couldn't have been caught. Conversion stays out of scope, but the contract check does not: all seven `.msgids` are now verified on every build. All six pass locally against the committed servers.

```
MIG-ABI-OK: task 3400-3442, 42 slots match the frozen contract
MIG-ABI-OK: vm_map 3800-3831, 31 slots
MIG-ABI-OK: host_priv 400-426, 26 slots
MIG-ABI-OK: mach_host 200-225, 25 slots
MIG-ABI-OK: mach_vm 4800-4820, 20 slots
MIG-ABI-OK: clock 1000-1003, 3 slots
```

PR #141, opened 2026-08-30
mach: generate the last five servers — MIG conversion complete (#125, #136)

### LINKED ISSUE #125: Restore MIG as the kernel RPC substrate — 7 subsystems, 27,455 lines of unregenerable generated C

### LINKED ISSUE #136: ipc_port_release_send() asserts IP_VALID, but modern migcom emits it unguarded — will panic on mach_vm conversion

pkgdemon:
Completes the MIG conversion. With this, **every Mach server the kernel builds is generated from a `.defs` in this tree** — no frozen 2015 server is left in the build.

Stacked on #140 (clock). Two commits: the #136 unblock, then the conversion.

| subsystem | msgids | slots |
|---|---|---|
| mach_host | 200–225 | 25 |
| host_priv | 400–426 | 26 |
| task | 3400–3442 | 42 |
| vm_map | 3800–3831 | 31 |
| mach_vm | 4800–4820 | 20 |

## Verified locally before pushing

Rather than discovering this in CI, I ran `ci/gen-mig-servers.sh` and `ci/verify-mig-abi.sh` on this machine against **all seven** subsystems:

```
mach_port  : MIG-ABI-OK: mach_port 3200-3236, 36 slots match the frozen contract
task       : MIG-ABI-OK: task 3400-3442, 42 slots match the frozen contract
vm_map     : MIG-ABI-OK: vm_map 3800-3831, 31 slots match the frozen contract
host_priv  : MIG-ABI-OK: host_priv 400-426, 26 slots match the frozen contract
mach_host  : MIG-ABI-OK: mach_host 200-225, 25 slots match the frozen contract
mach_vm    : MIG-ABI-OK: mach_vm 4800-4820, 20 slots match the frozen contract
clock      : MIG-ABI-OK: clock 1000-1003, 3 slots match the frozen contract
```

183 routine slots, base/end and every slot matching, tombstones included. That is the check that matters: `msgh_id` is base + index, so a misplaced `skip;` silently points a message at the **wrong function**.

## #136 had a wider blast radius than the ticket said

The ticket scoped #136 to mach_vm. It is not — generating the servers showed **16 unguarded `ipc_port_release_send()` calls** across four subsystems:

| subsystem | unguarded calls |
|---|---|
| host_priv | 5 |
| task | 6 |
| vm_map | 3 |
| mach_vm | 2 |

The frozen servers guard every one with `if (IP_VALID(p))`. Modern migcom emits them bare, because Apple moved the check **into the callee**; ours still had `assert(IP_VALID(port))`. Landing the conversion without fixing this would panic the kernel the first time any of those routines saw a null or dead port — an ordinary thing for a client to send.

The first commit fixes that the way Apple did. It is **behaviour-preserving**: `IP_VALID()` excludes both `IP_NULL` and `IP_DEAD`, so an early return does exactly what the caller-side guard did — only the side of the call the test lives on changes. Scoped to `ipc_port_release_send()` alone; `_sonce` and `_receive` keep their asserts, since the generated servers call neither.

`mach_host` needed no unblock at all (zero such calls).

## Why the struct diffs are expected

The `__Request__` structs are **not** byte-identical to the frozen ones. Every frozen request carries a `mach_msg_body_t msgh_body` between the "kernel processed data" markers with **zero descriptors in it** — an artifact of the 2015 MIG. Modern migcom omits it for non-complex requests.

`mach_port` is the control: I recovered its pre-conversion frozen server from git and it shows the *same* spurious field. It was converted first and its nine routines are verified working on hardware. Dropping the field is what the already-shipping subsystem does.

There is also no MIG client to disagree with — nextbsd-userland carries **no `.defs`** for any of these five, and libmach's entry points still fabricate success (#93). This changes what the kernel is *built from*, not what any existing caller sends.

## #137 unaffected

`vm_map` and `mach_vm` carry #137's two stale LP64 wire sizes. Those are reproduced deliberately: the `.defs` encodes the frozen numbers, so generated and frozen agree exactly. This PR neither fixes nor worsens them — #137 remains open on its own merits.

## Verification status

Local generation + ABI verification pass for all seven. Kernel **compilation** is CI's job (I can't cross-build FreeBSD here), and the smoke-test boot is the real gate. **Not verified on hardware** — the arm64 box is down (`panic: no init`).

Closes #125. Closes #136.


pkgdemon:
CI green on the second push.

```
kernel (amd64, amd64)                                    pass  6m46s
kernel (arm64, aarch64)                                  pass  7m35s
smoke-test (boot the self-assembled image with this kernel)  pass  1m53s
```

The smoke-test **boots** — 1m53s, in line with a normal run, not a timeout. Worth stating explicitly since a hang is the failure mode this repo's boot check would otherwise report as success.

## The first push failed, and the fix is in the third commit

`task_server.c` hit 19 implicit-declaration errors on `task_deallocate()` and `convert_task_to_port()`. `gen_one`'s kernel prologue splices a fixed set of six headers — everything mach_port and clock need — and two subsystems need one more apiece:

| subsystem | header | for |
|---|---|---|
| task | `sys/mach/task.h` | `task_deallocate`, `convert_task_to_port` |
| vm_map | `sys/mach/vm_types.h` | the `vm_*_t` wire types |

Determined by diffing frozen against generated includes for all five subsystems rather than by chasing the error message: `mach_host`, `host_priv` and `mach_vm` need nothing extra and get nothing. The splice is now checked, so a header that fails to land is a build failure at generation time rather than 19 errors later.

The change is confined to those two files — verified by regenerating everything before and after and diffing:

```
mach_port: byte-identical    host_priv: byte-identical
clock:     byte-identical    mach_vm:   byte-identical
mach_host: byte-identical
task:   +#include <sys/mach/task.h>       (one line)
vm_map: +#include <sys/mach/vm_types.h>   (one line)
```

That third commit also corrects the `<string.h>` comment, which claimed the generated servers call no str/mem function. True while mach_port and clock were the only ones generated; `mach_host`, `host_priv` and `task` call `memcpy()`/`strlen()` 2, 6 and 4 times. Dropping `<string.h>` is still right — the frozen servers make exactly the same calls the same number of times and resolve them through the kernel's own headers, since their `<string.h>` sits in the non-`_KERNEL` branch the kernel build never sees. Counted on both sides rather than assumed.

## Still unverified

Hardware. The arm64 box is down with `panic: no init` — a deploy accident (`tar -U` unlinked `/sbin/launchd`), unrelated to any code here. Recovery instructions are written up; the box never reached userland, so nothing about this change is implicated in it.

pkgdemon:
Superseded by #142 — GitHub auto-closed this when `conv/125-clock` was deleted by the #140 merge. Same commits, rebased onto main.

Issue #152, opened 2026-08-31
Stage 2 / C3: delete the fd backing (requires Stage 1)


pkgdemon:
Remove `ie_fp`, `mach_fileops`, and ~150 lines of copied FreeBSD fd internals.

**Move `mach_port_close`'s teardown body first — do not lose it.** It carries the real pset-destroy / `ipc_port_clear_receiver` / `ipc_port_destroy` / `ip_release` sequence.

### Delete the implicit fileport path
`ipc_object_copyin` opens by trying `ipc_entry_file_to_port`, which wraps **any ordinary fd** passed where a port name is expected into a special port with `CAP_ALL1` rights — an undeclared implicit fileport with no API and no guard. Structurally impossible after this stage. Userland's `fileport_*` are already `ENOSYS` stubs and every consumer handles the failure path, so delete rather than preserve.

### Add a per-space entry cap
`RLIMIT_NOFILE` implicitly caps port allocation today; afterwards nothing does. `ipc_entry_alloc_name` makes the index user-controlled, so an uncapped grow-to-cover is a memory-amplification vector. Refuse above a threshold rather than growing.

### Known losses
Ports disappear from `fstat(2)`, `procstat -f` and `lsof`; `close(2)` on a port name stops destroying the port; port rights can no longer be sent via `SCM_RIGHTS`. C6 restores the introspection.

Blocked on: a dedicated sweep for userland code using a port name as an fd. `libmach` is clean; launchd, libdispatch, CoreFoundation, configd, syslogd and mDNSResponder have **not** been audited.

Part of #145. Plan: https://pkgdemon.github.io/nextbsd-mach-namespace-separation-plan.html

pkgdemon:
## Blocking sweep done: no userland code uses a port name as an fd

This ticket was blocked on a dedicated sweep, since `libmach` alone had been checked. Swept the whole userland tree.

**`SCM_RIGHTS` senders** — four files: `libxpc/msgio.c`, `launchd/liblaunch/liblaunch.c`, `mDNSResponder/mDNSShared/uds_daemon.c`, `mDNSResponder/mDNSShared/dnssd_clientstub.c`. All pass genuine `int` file descriptors:

```c
/* libxpc/msgio.c */          PJDLOG_ASSERT(fd >= 0);  bcopy(&fd, CMSG_DATA(cmsg), sizeof(fd));
[...]

Issue #145, opened 2026-08-31
Mach namespace separation: stop port names being file descriptors


pkgdemon:
Mach port names are file descriptors in this port, allocated by `kern_fdalloc(td, 16, &fd)` (`ipc_entry.c:453`), and the generation counter is compiled out (`sys/mach/port.h:219`, `MACH_PORT_GEN(name)` is `(0)`). A name freed by one RPC is handed back bit-identical to the next, and nothing can detect that a cached name now denotes a different object. `ipc_object_copyin()` then wraps any non-Mach fd into a port with `ip_receiver` set and nobody receiving, so a send succeeds and the message is swallowed.

XNU never had this: a port name is an `ipc_space` table index plus a generation, and the Mach namespace is separate from the BSD fd namespace. This umbrella tracks undoing that deviation.

## Measured on arm64 hardware

- Permanent wedge, not slowness — a 90s timeout expires unchanged
- Bimodal: sub-second success or never
- Client and server both blocked in `ipc_mqueue_receive` simultaneously
- 40–60% of sends wedge from cold boot, falling to 0 after ~70–80 cumulative sends
- 10 minutes idle with zero sends changes nothing — the driver is allocation, not time
- Restarting **either** daemon reinstates the failures

## Two stages

**Stage 1** — make the kqueue filter stop depending on fds. `machport_filtops` has `.f_isfd = 1` (`ipc_pset.c:537`), so kqueue calls `fget(kev.ident)` before attach, and libmach registers with the port *name* as ident. Until this changes, generationed names are impossible.

**Stage 2** — names become table indices with generations.

## Not claimed

These are **expected** to fix the wedge; that is not demonstrated. Four predicted fixes during this investigation (#126, #129, #130, and kernel #144) each addressed a real defect and none moved the measured rate. Treat as unproven until the acceptance measurement says otherwise.

Full plan, change maps and risks: https://pkgdemon.github.io/nextbsd-mach-namespace-separation-plan.html

pkgdemon:
## Ticket index

**Stage 1 — de-fd the kqueue filter**
- #146 prereq: `ipc_pset_alloc_name` never initialises `ips_note` / `ips_note_lock` (land first, alone)
- #147 make EVFILT_MACHPORT resolve through `ipc_space`

**Stage 2 — names become table indices**
- #148 C0: separate the reverse hash off `is_table` (prep, no semantic change)
[...]

PR #161, opened 2026-08-31
mach/ipc: refuse to resolve a generationed name as a file (#152, partial)

pkgdemon:
Part of #145. Follows #160, and closes a gap I found by re-reading my own patch.

## The gap

`ipc_object_copyin()` calls `ipc_entry_file_to_port()` **before** `ipc_right_lookup()`. That function `fget`s the name's index half and, if it denotes an ordinary file descriptor, wraps it in a synthetic port with `ip_receiver` set and **nobody receiving on it** — a send to which succeeds and is never delivered.

Because it runs before the lookup, **it bypasses the generation check added in #160.**

| stale name lands on | today | with #160 | with this |
|---|---|---|---|
| another **Mach** entry | wedge | rejected (lookup + gen check) | rejected |
| an ordinary **fd** | wedge | **still wedges** | rejected |

That second row is specifically the observed failure mode in #145 — client's send succeeds, server never sees the message, client blocks forever. #160 alone would only have closed half of it.

## The guard

A legitimately-obtained port name always carries a **non-zero** generation: `ipc_entry_get()` stamps `IE_BITS_NEW_GEN()` on every allocation and the first value is `IE_BITS_GEN_ONE`, never 0. A bare fd number passed where a port name is expected has none.

So a non-zero generation means "this is a Mach port name" — and a Mach port name must never be resolved as a file.

## Safety

- Callers unaffected: `ipc_object_copyin()` treats any non-`KERN_SUCCESS` as "not a file, continue with the normal port path".
- No legitimate consumer lost: the implicit fileport has no API and no guard, userland's `fileport_*` are `ENOSYS` stubs, and the sweep recorded on #152 found nothing in userland passing an fd where a port name is expected.

Deleting the path entirely is #152 proper; this is the minimal guard that makes it safe under generations, and it is the half that matters for the wedge.

### DIFF
--- a/src-overlay/sys/compat/mach/ipc/ipc_entry.c
+++ b/src-overlay/sys/compat/mach/ipc/ipc_entry.c
@@ -376,6 +376,33 @@ ipc_entry_file_to_port(ipc_space_t space, mach_port_name_t name, ipc_object_t *o
 	if (curthread->td_proc->p_fd == NULL)
 		return (KERN_INVALID_ARGUMENT);
 
+	/*
+	 * Refuse a name that carries a generation.
+	 *
+	 * This function is the implicit-fileport path: ipc_object_copyin()
+	 * calls it BEFORE ipc_right_lookup(), and if the name's index half
+	 * happens to denote an ordinary file descriptor it wraps that fd in a
+	 * synthetic port with ip_receiver set and nobody receiving on it. A
+	 * send to such a port succeeds and the message is never delivered.
+	 *
+	 * Because it runs before the lookup, it also bypasses the generation
+	 * check in ipc_entry_lookup(). So a STALE Mach port name whose index
+	 * has since been recycled to a vnode or socket was still being
+	 * silently converted into a receiverless port -- the generation was
+	 * stamped and validated, but this path never consulted it. That is
+	 * half of the aliasing the generation counter was added to catch, and
+	 * it is the half that produces the observed wedge (#145).
+	 *
+	 * A legitimately-obtained port name always has a non-zero generation:
+	 * ipc_entry_get() stamps IE_BITS_NEW_GEN() on every allocation and
+	 * the first value is IE_BITS_GEN_ONE, never 0. A bare file descriptor
+	 * number passed where a port name is expected has none. So a non-zero
+	 * generation means "this is a Mach port name" -- and a Mach port name
+	 * must never be resolved as a file.
+	 */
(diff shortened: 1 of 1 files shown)


pkgdemon:
## Hold — #160 is being reverted, so this guard is inert

#160 (the generation counter) **passed CI's boot smoke-test and then failed to boot on the arm64 box.** The machine had to be recovered via LiveCD. It is being reverted in #162.

That makes this PR a no-op rather than wrong: without generations, `MACH_PORT_GEN(name)` is always 0, so the new `if (MACH_PORT_GEN(name) != 0) return KERN_INVALID_NAME;` never fires. Harmless, but it protects against nothing until generations are back.

**This should land together with the generation retry, not before it.** The reasoning behind it is unaffected and still correct — `ipc_object_copyin()` calls `ipc_entry_file_to_port()` before `ipc_right_lookup()`, bypassing the generation check, so a stale name whose index now denotes an ordinary fd still gets wrapped in a receiverless port. That is the half of the aliasing that produces the wedge, and #160 alone would not have closed it.

Leaving the PR open rather than closing it, so the analysis and the patch stay attached to the retry.

pkgdemon:
Closing: this cannot fire on current main.

The guard depends on `ipc_entry_get()` stamping `IE_BITS_NEW_GEN()`, but #160 — which added generations — was merged and then **reverted** in `08f42c0` (PR #162) because it passed CI and left the box unbootable. Verified against main: `IE_BITS_NEW_GEN` and `IE_BITS_GEN_ONE` are used nowhere, and there is no live `MACH_PORT_GEN`. Every name has generation 0, so the check never triggers.

Its stated motivation is also superseded. This PR attributes the #145 wedge to the implicit fileport path — *"client's send succeeds, server never sees the message, client blocks forever"*. That was a reasonable read at the time, but the wedge turned out to be `EV_CLEAR` on libmach's `EVFILT_MACHPORT` readiness knote, fixed in nextbsd-userland `f20893d` (22/80 → 80/80). The fileport path was not involved.

The underlying hazard is still real — an audit of #145 independently flagged it: `ipc_object_copyin()` tries `ipc_entry_file_to_port()` first, wrapping any ordinary fd in a receiverless port so the send succeeds and the message is swallowed. But the right fix is to **delete** the path (~90 lines, no legitimate producer: `fileport_makeport` is an ENOSYS stub), not to guard it with a check that depends on a reverted feature.

That deletion is prepared on `fix/delete-implicit-fileport` and can be opened when kernel work resumes.

Issue #112, opened 2026-08-25
rpi5: no thermal sensor or fan control — the Active Cooler never spins


pkgdemon:
NextBSD has no thermal sensor driver, no cpufreq/DVFS, and no thermal-zone framework on BCM2712. Most of that turns out not to matter. One part does, and it is Pi 5 B specific.

## What is *not* a problem

Thermal **protection** on this SoC is firmware-side. The VPU owns clocks and voltage and throttles on its own (soft limit ~80 °C, hard ~85 °C). Linux's cpufreq on Pi is a *client* of that firmware over the mailbox, not the thing enforcing the limit — so having no DVFS does not remove a safety net the OS never held. No board is at risk of damage, and FreeBSD's arm64 idle loop uses WFI, so idle power is not pathological either.

## What *is* a problem — the fan

The vendor device tree we ship drives the Active Cooler entirely from the **Linux** thermal framework. From `bcm2712-rpi-5-b.dtb` (raspberrypi/firmware 1.20250430), read straight out of the blob:

```
/cooling_fan     compatible = "pwm-fan"    status = "disabled"
                 cooling-levels = <0 75 125 175 250>    # level 0 = off
                 #cooling-cells = <2>

/thermal-zones/cpu-thermal
   thermal-sensors = <&thermal>           # brcm,bcm2711-thermal @ 0x107d542000
   coefficients = <(-550) 450000>
   trips:  cpu-tepid  50000 active    cpu-hot   67500 active
           cpu-warm   60000 active    cpu-vhot  75000 active
           cpu-crit  110000 critical
   cooling-maps: tepid/warm/hot/vhot/melt -> &cooling_fan, levels 1..4
```

`pwm-fan`, `thermal-zones`, `cooling-maps` and `cooling-levels` are Linux bindings. NextBSD reads the same blob with its own drivers and implements none of them, so every one of those nodes is inert: **the fan never spins**, and the 110 °C critical trip (an orderly shutdown under Linux) never fires either — though firmware throttling bites long before that.

Same class of problem as the `dtoverlay=` note in nextbsd's `build.sh`: vendor device tree written against Linux driver bindings, read by drivers that do not implement them.

## Per board

- **Pi 500 / 500+** — no fan at all. Passively cooled through the heatsink plate in the keyboard chassis. Unaffected, which is exactly why this has not surfaced: it is the only board that has ever been booted.
[...]

PR #171, opened 2026-09-02
rpi5: fix the BCM2712 mailbox DMA clamp (black screen), and read SoC temperature/throttling

pkgdemon:
Two changes for the Pi 5 family, shipped together so testers exercise **one** kernel and report both halves from the same boot. Fixes the video regression first, then adds temperature visibility.

Refs #112, and nextbsd#422 (call for testing).

---

## 1. `0030` — clamp the BCM2712 mailbox DMA to what the VideoCore can actually reach

**This is a fix to code already in `main`, and it is the more urgent half.**

Two ceilings apply to a mailbox property buffer. `0027` accounted for the looser one:

> The mailbox property interface carries the buffer address in a single 32-bit word, so a buffer handed to the VideoCore has to live below 4GB.

True, and **not sufficient**. The VideoCore's reach into SDRAM is tighter than the register is wide — which is why `brcm,bcm2711` and `brcm,bcm2838` both clamp to `BCM2838_PERIPH_MAXADDR` (`0x3fffffff`) rather than the full 32-bit range. 2712 is no different; the clamp was missed.

### Measured

A Pi 500+ with 16GB. The firmware hands physical memory over in three chunks:

```
0x00000000080000 - 0x0000000001ffff      1.5 MB
0x000000013ca000 - 0x0000003f3fffff    1040 MB
0x00000040232000 - 0x000003ea657fff   15741 MB
```

Instrumenting `bcm2835_mbox_property()` to log the buffer address and the VideoCore's response code:

```
req tag=0x00030006 phys=0x0019f000  ->  code=0x80000000 (SUCCESS)   chunk 1
req tag=0x00040003 phys=0x013ca000  ->  code=0x80000000 (SUCCESS)   chunk 2
req tag=0x00040005 phys=0x40232000  ->  code=0x00000000 (FAILS)     chunk 3
```

Response code `0x00000000` is the *request* code — the buffer came back untouched, because the VideoCore never wrote to it. Every call landing in the low two chunks succeeded; the one landing in the third did not. The boundary sits between `0x3f3fffff` and `0x40232000`, which `BCM2838_PERIPH_MAXADDR` already covers.

**Stated honestly:** 1GB is *sufficient* and measured, not proven to be the exact ceiling. The lowest failing address available to test was the start of the third chunk, so `0x40000000`–`0x40232000` is unprobed. The value matches the two SoCs whose limit is known rather than sitting at the edge of what was measured.

### Why it matters beyond this PR

The failure is **silent** and depends on where the allocator happens to place a buffer, so it moves with the memory map. `bcm2835_fbd` makes two property calls during attach — `GET_PHYSICAL_W_H` then `GET_DEPTH` — and a layout that puts the second above 1GB kills the console, leaving nothing in dmesg but `mbox response error`, which until now did not even print the code it rejected.

Changing `framebuffer_depth` in `config.txt` resizes the firmware's reservation, shifts the chunk boundaries, and changes which calls survive.

**That is very likely the real mechanism behind the per-board `framebuffer_depth` workarounds in nextbsd#422** — one board needing `32`, another needing `24`, a third getting no display either way. Three memory maps, not three pixel formats. It would also explain why the "24bpp packs three bytes where the driver writes four" theory predicted a *smeared* console while the actual reports were *no picture at all*.

**Unverified on the two boards that are not mine.** That is part of what this build is for.

---

## 2. `0031` / `0032` — temperature and throttling

Unchanged from the previous revision of this PR.

`0031` builds two **existing** files on `soc_brcm_bcm2712`:

| file | what it gives |
|---|---|
| `bcm2835_firmware.c` | attaches to `/soc/firmware`, exports `bcm2835_firmware_property()`, adds the `dev.bcm2835_firmware.0` sysctl node |
| `bcm2835_vcio.c` | `/dev/vcio` — what `sysutils/raspberrypi-userland`'s `vcgencmd` opens |

Neither needs porting; both were gated to the Pi 3 and Pi 4 SoCs only. `0027` already brought the property mailbox up on this SoC and proved it.

`0032` adds two read-only sysctls beside the existing `revision`:

```
dev.bcm2835_firmware.0.temperature      # "IK" — sysctl(8) prints degrees C
dev.bcm2835_firmware.0.throttled        # raw flag word
```

`throttled` is the more useful of the two: bits 0–3 are live state (undervolt / frequency capped / throttled / soft temperature limit), bits 16–19 latch the same conditions until reboot. Neither has a write path — the VPU owns the clock, the rails and the thermal limits. This closes the gap nextbsd#422 documents as *"NextBSD cannot currently read the SoC temperature either."*

### Verified on hardware

```
dev.bcm2835_firmware.0.temperature: 36.6C
dev.bcm2835_firmware.0.throttled:   0
dev.bcm2835_firmware.0.revision:    1762364238
```

---

## Ordering

`0030` first, deliberately. It is a standalone fix to code already in `main` and can be cherry-picked without the feature; `0031`/`0032` depend on nothing in it.

## Not included

- **`bcm2835_cpufreq.c`.** Looks like the same gate flip; is not. It *sets* clocks and voltages on a SoC whose firmware owns both, over a tag set not verified to match 2837/2838.
- **The fan.** RP1 PWM1 channel 3, behind PCIe — needs a `raspberrypi,rp1-pwm` driver. That is #172.

## Caveat: boards whose firmware refuses to change depth

`0034` makes those boards usable rather than perfect, and the distinction matters for anyone testing.

Some firmware accepts `SET_DEPTH` as a *setting* without reallocating the framebuffer. Verified through `/dev/vcio` on a Pi 500+ with no `framebuffer_depth` in `config.txt`, after `SET_DEPTH` reported success:

```
GET_DEPTH    -> 32
GET_PITCH    -> 3840        (1920 * 2: a 16bpp stride)
GET_PHYS_W_H -> 1920 x 1080
allocation   -> 4147200 bytes = 1920 * 1080 * 2
```

Every field except the depth describes a 16bpp buffer, so `0034` renders at 16bpp. The same board needs `framebuffer_depth` under Linux too, so this is firmware behaviour rather than something a driver can ask its way around.

Measured outcome on such a board:

| | before | with `0030` + `0034` |
|---|---|---|
| console | black screen, or garbled | **works**, at 16bpp |
| desktop (X) | — | **does not start** |

X dies because `scfb` segfaults on a 16bpp framebuffer — nextbsd#425, filed separately with the evidence. Verified as depth-only: same kernel, same board, same packages, X works at 32bpp and segfaults at 16bpp, and the `mesa-dri` / `xauth` / `card0` errors in that log are present and harmless at 32bpp too.

**So if your board lands on the 16bpp fallback and you want a desktop, put `framebuffer_depth=32` back in `config.txt`.** The console works either way, which it did not before.

Strictly better in every case and worse in none — but "console yes, desktop no" is a real limitation, not a footnote.

## Testing

Full 32-patch series applies cleanly to `releng/15.1` at `88e7371d9dc`. CI builds `NEXTBSD` (amd64 + arm64) and `NEXTBSD-RPI5` + `kernel8.img`.

**If you are testing this, please report both halves — video and temperature are equally the point.**

Flash `rpi5-kernel8-arm64` from this PR's run as `kernel8.img`, and **first remove any `framebuffer_depth` workaround** from `config.txt` so the shipped default is what gets exercised:

```sh
# 1. video — does the console come up with the stock config.txt?
#    (report your board, and what depth you previously needed, if any)

# 2. temperature
sysctl dev.bcm2835_firmware
printf '%#x\n' $(sysctl -n dev.bcm2835_firmware.0.throttled)   # 0x0 on a healthy PSU
```

If `dev.bcm2835_firmware.0` is missing entirely, `dmesg | grep -i firmware` separates "driver never attached" from "the 2712 mailbox rejected the tag". If video is still black, `dmesg | grep -iE 'fb0|mbox'` over SSH is the useful thing — the board still boots and takes SSH with no display.

### DIFF
--- a/patches/0030-bcm2835_vcbus-clamp-bcm2712-mailbox-dma-to-videocore-reach.patch
+++ b/patches/0030-bcm2835_vcbus-clamp-bcm2712-mailbox-dma-to-videocore-reach.patch
@@ -1,0 +1,124 @@
+From a33c931fb92dd919e6feffc13bb8adab0bdebedd Mon Sep 17 00:00:00 2001
+From: pkgdemon <jpm820@gmail.com>
+Date: Wed, 2 Sep 2026 22:55:21 -0500
+Subject: [PATCH] bcm2835_vcbus: clamp BCM2712 mailbox DMA to the VideoCore's
+ reach
+
+Two ceilings apply to a mailbox property buffer, and only the looser one
+was accounted for. 0027 set busdma_lowaddr to BUS_SPACE_MAXADDR_32BIT with
+the reasoning that the property interface carries the buffer address in a
+single 32-bit word. That is true, and not sufficient: the VideoCore's reach
+into SDRAM is tighter than the register is wide, which is why 2711 and 2838
+both clamp to BCM2838_PERIPH_MAXADDR instead. 2712 is no different; the
+clamp was simply missed.
+
+Measured on a Pi 500+ with 16GB. The firmware hands physical memory over in
+three chunks:
+
+  0x00000000080000 - 0x0000000001ffff      1.5 MB
+  0x000000013ca000 - 0x0000003f3fffff    1040 MB
+  0x00000040232000 - 0x000003ea657fff   15741 MB
+
+Instrumenting bcm2835_mbox_property() to log the buffer address and the
+VideoCore's response code:
+
+  req tag=0x00030006 phys=0x0019f000  -> code=0x80000000 (SUCCESS)  chunk 1
+  req tag=0x00040003 phys=0x013ca000  -> code=0x80000000 (SUCCESS)  chunk 2
+  req tag=0x00040005 phys=0x40232000  -> code=0x00000000 (FAILS)    chunk 3
+
+Response code 0x00000000 is the request code: the buffer came back
+untouched because the VideoCore never wrote to it. Every call landing in
+the low two chunks succeeded; the one landing in the third did not. The
+boundary sits between 0x3f3fffff and 0x40232000, which
+BCM2838_PERIPH_MAXADDR already covers.
+
+Stated honestly: 1GB is sufficient and measured, not proven to be the exact
+ceiling. The lowest failing address available to test was the start of the
+third chunk, so 0x40000000-0x40232000 is unprobed. The value matches the
+two SoCs whose limit is known rather than sitting at the edge of what was
+measured.
+
+What makes this worth fixing ahead of anything else: the failure is silent
+and depends on where the allocator happens to place a buffer, so it moves
+with the memory map. bcm2835_fbd makes two property calls during attach --
+GET_PHYSICAL_W_H then GET_DEPTH -- and a layout that puts the second one
+above 1GB kills the console with nothing in dmesg but "mbox response
+error", which until now did not even print the code it rejected.
+
+Changing framebuffer_depth in config.txt resizes the firmware's
+reservation, shifts the chunk boundaries, and changes which calls survive.
+That is very likely the real mechanism behind the per-board
+framebuffer_depth workarounds in the Pi 5 call for testing -- where one
+board needs 32, another needs 24, and a third gets no display either way.
+Those are three different memory maps, not three different pixel formats.
+Unverified on the two boards that are not mine.
+
+Refs nextbsd-kernel#171, and nextbsd#422 (call for testing).
+---
+ sys/arm/broadcom/bcm2835/bcm2835_vcbus.c | 49 ++++++++++++++++++++----
+ 1 file changed, 42 insertions(+), 7 deletions(-)
+
+diff --git a/sys/arm/broadcom/bcm2835/bcm2835_vcbus.c b/sys/arm/broadcom/bcm2835/bcm2835_vcbus.c
+index 565d9f336c0..bcb9079e43c 100644
+--- a/sys/arm/broadcom/bcm2835/bcm2835_vcbus.c
++++ b/sys/arm/broadcom/bcm2835/bcm2835_vcbus.c
+@@ -211,14 +211,49 @@ static struct bcm283x_memory_soc_cfg {
+ 		.memmap = bcm2712_memmap,
+ 		.soc_compat = "brcm,bcm2712",
+ 		/*
+-		 * The mailbox property interface carries the buffer address in
+-		 * a single 32-bit word, so a buffer handed to the VideoCore has
+-		 * to live below 4GB whatever the interconnect can reach. That
+-		 * is a property of the protocol rather than of this SoC, and it
+-		 * is why this is not the full identity range the dma-ranges
+-		 * would otherwise allow.
++		 * Two separate ceilings apply here, and only the looser one was
++		 * accounted for originally.
++		 *
++		 * The mailbox property interface carries the buffer address in a
++		 * single 32-bit word, so a buffer handed to the VideoCore must
++		 * live below 4GB. True, and a property of the protocol rather
++		 * than of this SoC -- but not sufficient. The VideoCore's own
++		 * reach into SDRAM is tighter than the register is wide, which
++		 * is why 2711 and 2838 above both clamp to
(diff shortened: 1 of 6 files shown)


pkgdemon:
## How to test this kernel

You are replacing **one file** — `kernel8.img` on the FAT32 boot partition. Nothing else changes, and the old kernel stays on the partition so you can put it back.

### 1. Get the kernel

Open the build for this PR and download the **`rpi5-kernel8-arm64`** artifact:

**https://github.com/nextbsd/nextbsd-kernel/actions/runs/33716473785**

Scroll to **Artifacts** at the bottom of that page. You need to be signed in to GitHub to download it — that is a GitHub restriction on Actions artifacts, not a choice here. It expires 2026-12-02.

Or with the `gh` CLI:

```sh
gh run download 33716473785 --repo nextbsd/nextbsd-kernel -n rpi5-kernel8-arm64
```

Either way you get a zip containing a single file, `kernel8.img` (~15 MB). Unzip it.

Sanity-check you have the right one before copying it anywhere:

```sh
strings -a kernel8.img | grep -m1 "NextBSD Kernel Version"
# expect: NextBSD Kernel Version 20260903-045603
```

### 2a. If you boot from the USB stick

Plug the stick into any computer. The first partition is FAT32, labelled **NEXTBSD**, and every desktop OS mounts it automatically — macOS at `/Volumes/NEXTBSD`, Linux usually under `/media/…`, Windows as a drive letter.

```sh
cd /Volumes/NEXTBSD          # or wherever it mounted

cp kernel8.img kernel8.bak   # keep the working one — this is your way back
cp /path/to/new/kernel8.img kernel8.img
```

Then **eject properly** rather than just unplugging — the copy sits in the write cache otherwise and you will boot a truncated kernel. `Eject` in Finder, the eject icon on Linux, "Safely Remove" on Windows. Or from a terminal: `sync && diskutil unmount /Volumes/NEXTBSD` (macOS) / `sync && umount /media/…` (Linux).

### 2b. If you already installed to NVMe

The boot partition is on the NVMe, not the stick, so do it from the running system — SSH in or use the console:

```sh
mkdir -p /mnt/efi
mount -t msdosfs /dev/nda0s1 /mnt/efi        # check `gpart show` if nda0 is not your disk

cp -p /mnt/efi/kernel8.img /mnt/efi/kernel8.bak
cat /path/to/new/kernel8.img > /mnt/efi/kernel8.img
chmod 755 /mnt/efi/kernel8.img

sync; umount /mnt/efi
```

### 3. Before you reboot — remove any `framebuffer_depth` workaround

**This matters more than anything else here.** If you previously changed `framebuffer_depth` to get a picture, put it back to the shipped default:

```
framebuffer_depth=32
```

The whole point of this build is to find out whether the stock setting now works. Testing with your workaround still in place tells us nothing.

### 4. Reboot, and report both

```sh
uname -v                       # expect 20260903-045603

sysctl dev.bcm2835_firmware    # temperature, throttled, revision
printf '%#x\n' $(sysctl -n dev.bcm2835_firmware.0.throttled)
```

**If your console comes up but the desktop does not start**, check `dmesg` for:

```
fb0: firmware claims 32bpp but allocated a NNNN-byte stride; using 16bpp
```

That means your firmware would not restructure the framebuffer, so the console fell back to 16bpp — a known limitation, not a new bug. The console working at all is the improvement; X cannot yet cope with 16bpp (nextbsd#425). Put `framebuffer_depth=32` back in `config.txt` if you want the desktop, and **please report that you hit it** — that is a data point worth having.

Please report **both** halves — video and temperature are equally the point of this build:

1. **Video** — did the console come up, at stock `framebuffer_depth=32`? Which board, and what depth did you previously need?
2. **Temperature** — what does `sysctl dev.bcm2835_firmware` print?

Positive results are as useful as negative ones. "Pi 5 B, worked at 32, previously needed 24" is exactly the data point that settles this.

### If it still does not work

The board boots and takes SSH even with no display, so:

```sh
dmesg | grep -iE 'fb0|mbox'
```

and paste that. `mbox response error` now prints the response code it rejected, which it never used to.

### Going back

```sh
cp kernel8.bak kernel8.img     # from wherever you made the backup
```

Eject or `sync; umount` again, and reboot.

PR #172, opened 2026-09-02
rpi5: drive the Pi 5 B Active Cooler — rp1_clk, RP1 pinctrl, rp1_pwm

pkgdemon:
Fan half of #112. Stacked on #171 — both append to `patches/series`, so this
targets that branch and GitHub will retarget it to `main` when #171 merges.

Together these make the Active Cooler reachable from `pwm(8)`. The
temperature-driven policy that should step it automatically is deliberately
**not** here — see "What's still missing".

## The fan needed more than a PWM driver

From `bcm2712-rpi-5-b.dts:396`:

```
pwms = <&rp1_pwm1 3 41566 PWM_POLARITY_INVERTED>;
```

RP1 PWM1 channel 3 — behind PCIe on the southbridge. Two things stand between
that block and a spinning fan, and both are silent when missing.

### `0031` — `rp1_clk`, peripheral clock gates

RP1's leaf clocks are off at reset. The vendor driver that manages them is
2500 lines of PLL, mux and divider tree; this is the small part of it — the
clocks that take a fractional divider straight off the 50MHz crystal, which is
already running (everything on RP1 that works today works because it is).

One operation, *"run clock N at F Hz"*, and `ENOENT` for anything else. **The
refusal is the point.** A clock whose parent is a PLL cannot be added to the
table without also starting that PLL, and a divider programmed off a dead PLL
doesn't fail — it yields a peripheral that reads and writes normally and never
ticks.

Parent selection is the subtle bit: for these clocks the CTRL register's `SRC`
field is unused (the vendor table gives them `clk_src_mask = 0`, so its
`set_parent` writes nothing there) and selection is entirely by `AUXSRC`,
indexing that clock's own auxiliary parent list. `xosc` is at index 2 in both
PWM lists.

### `0032` — `rp1_gpio`, `fdt_pinctrl` provider

`rp1_gpio` was written with muxing left out on the grounds that the firmware
had already assigned every pin used so far — true for `phy-reset-gpios` and
`gpioled`. It stops being true at the fan:

```
&rp1_pwm1 { pinctrl-0 = <&rp1_pwm1_gpio45>; };
rp1_pwm1_gpio45 { function = "pwm1"; pins = "gpio45"; bias-pull-down; };
```

Linux applies that at probe. Nothing in the firmware does.

Two things easy to get wrong, and silent when wrong:

- **Muxing releases the overrides, not just FUNCSEL.** OUT/OE/IN can each be
  forced independently of the selected function. A pin left with `OEOVER`
  forced from an earlier GPIO configuration keeps that direction whatever
  FUNCSEL says, and the peripheral drives nothing.
- **The pad is handed over before the function is selected** — input buffer on,
  output driver enabled, *then* FUNCSEL. The other order briefly points the new
  function at a disabled output stage.

The per-pin function table is transcribed mechanically from the vendor
`pinctrl-rp1.c` (a script, not by hand) and must stay identical to it: the index
into a row **is** the FUNCSEL value and none of it is derivable. The `5` that
appears throughout at `"gpio"` is the `RP1_CTRL_FUNCSEL_RIO` this file already
writes, which is a free check that the transcription is aligned.

### `0033` — `rp1_pwm`

A `pwmbus(4)` provider, so `pwm(8)` reaches it with no new userland. The block
is a counter per channel compared against RANGE (period) and DUTY (on-time).
Two non-obvious properties:

- **Writes don't take effect when made.** Every register is shadowed and
  latched together on `SET_UPDATE` to `GLOBAL_CTRL`, so each change ends in one
  commit — which is also what stops a channel emitting a half-updated period.
- **The clock isn't ours.** Attach fails loudly if `rp1_clk` won't start it,
  rather than leaving a driver that looks attached and does nothing.

ns→ticks is done in one step rather than the vendor's two (it computes whole
nanoseconds-per-tick first, then divides). At 50MHz a tick is exactly 20ns and
they agree; at a rate that doesn't divide 1e9 evenly the vendor's first
rounding is silently lost.

## None of this is proven on hardware

**And it cannot be proven on the board we have.** A Pi 500+ has no fan header,
and its DTB leaves both `cooling_fan` and `rp1_pwm1` `status = "disabled"` —
`rp1_pwm` will not even attach there. Each driver says so in its own header
comment, not just here.

This needs a **Pi 5 B with an Active Cooler fitted**. Driven by hand:

```sh
pwm -f pwmc0.3 -p 41566 -d 50% -E    # ~50% — should be audible
pwm -f pwmc0.3 -p 41566 -d 100% -E   # full
pwm -f pwmc0.3 -D                    # off
pwm -f pwmc0.3 -C                    # show what the driver thinks it is doing
```

`pwmbus` adds one `pwmc(4)` child per channel, so the fan on channel 3 of the
first (and, with `rp1_pwm0` disabled, only) controller is `/dev/pwm/pwmc0.3`.
`pwm -f` takes the name relative to `/dev/pwm/`.

The DT says `PWM_POLARITY_INVERTED`, so if the fan runs *backwards* from the
duty cycle, add `-I`.

### One open question the code is arranged to answer

Whether the Pi 5 B firmware has already muxed GPIO45 before we get there.
Nobody here can find out. `0032` logs the FUNCSEL it found *before* overwriting
it, at `bootverbose`, so the first tester's dmesg settles it:

```
gpio0: gpio45 funcsel 31 -> 0 (pwm1)     # 31 = NULL, firmware left it alone
gpio0: gpio45 funcsel 0 -> 0 (pwm1)      # firmware had already done it
```

## What's still missing

The **policy**. `thermal-zones` / `cooling-maps` / `cooling-levels` are Linux
bindings and nothing here reads them. Once the fan can be made to spin by hand,
the policy is a small consumer that reads
`dev.bcm2835_firmware.0.temperature` (from #171) and steps the duty cycle
through the DT's `cooling-levels = <0 75 125 175 250>` against the trip points
in `cpu-thermal`. That's a follow-up, and it wants a fan that is known to work
underneath it first.

Verified: the full 33-patch series applies cleanly to `releng/15.1` at
`88e7371d9dc`.

Refs #112.

### DIFF
--- a/patches/0035-rp1_clk-peripheral-clock-gates-for-rp1.patch
+++ b/patches/0035-rp1_clk-peripheral-clock-gates-for-rp1.patch
@@ -1,0 +1,416 @@
+From 22ec8d0650fdc43fb8b689d34e8734f479fd7a26 Mon Sep 17 00:00:00 2001
+From: pkgdemon <jpm820@gmail.com>
+Date: Wed, 2 Sep 2026 12:13:27 -0500
+Subject: [PATCH 1/3] rp1_clk: peripheral clock gates for RP1
+
+Every leaf clock in RP1 is off at reset, and the vendor driver that turns
+them on is 2500 lines of PLL, mux and divider tree. This is the small part
+of that: the clocks that take a fractional divider straight off the 50MHz
+crystal, which is already running -- everything on RP1 that works today
+works because it is.
+
+So the interface is one operation, "run clock N at F Hz", for the clocks
+in the table, and ENOENT for everything else. That refusal is the point.
+A clock whose parent is a PLL cannot be added to the table without also
+bringing that PLL up, and a divider programmed off a dead PLL does not
+fail -- it produces a peripheral that reads and writes normally and never
+ticks.
+
+Two clocks to begin with, PWM0 and PWM1, because the Pi 5 B's Active
+Cooler is on pwm1 (nextbsd-kernel#112). The ids are the vendor binding's
+(include/dt-bindings/clock/rp1.h) because they arrive as the second cell
+of a `clocks = <&rp1_clocks N>` phandle out of the device tree we are
+handed; they are ABI, not ours to number.
+
+Parent selection is the one part worth spelling out. For these clocks the
+CTRL register's SRC field is unused -- the vendor's table gives them
+clk_src_mask = 0, so its set_parent writes nothing there -- and selection
+is entirely by the AUXSRC field, indexing that clock's own auxiliary
+parent list. xosc is at index 2 in both PWM lists. The divider is 16.16
+fixed point across two registers, and the arithmetic here is
+rp1_clock_choose_div() from the vendor driver rather than re-derived.
+
+Register offsets are from clk-rp1.c.
+
+The driver attaches at BUS_PASS_BUS, ahead of rp1_gpio (BUS_PASS_INTERRUPT)
+and rp1_pwm (BUS_PASS_DEFAULT), so consumers find it without anyone having
+to encode an explicit dependency.
+
+UNPROVEN ON HARDWARE: written against the vendor driver and the device
+tree, not against a board. The only Pi 5-family machine available is a
+Pi 500+, whose DTB leaves both PWM blocks disabled.
+
+Refs nextbsd-kernel#112.
+---
+ sys/arm/broadcom/bcm2835/rp1_clk.c | 288 +++++++++++++++++++++++++++++
+ sys/arm/broadcom/bcm2835/rp1_clk.h |  50 +++++
+ sys/conf/files.arm64               |   1 +
+ 3 files changed, 339 insertions(+)
+ create mode 100644 sys/arm/broadcom/bcm2835/rp1_clk.c
+ create mode 100644 sys/arm/broadcom/bcm2835/rp1_clk.h
+
+diff --git a/sys/arm/broadcom/bcm2835/rp1_clk.c b/sys/arm/broadcom/bcm2835/rp1_clk.c
+new file mode 100644
+index 00000000000..1c052b693c4
+--- /dev/null
++++ b/sys/arm/broadcom/bcm2835/rp1_clk.c
+@@ -0,0 +1,288 @@
++/*-
++ * SPDX-License-Identifier: BSD-2-Clause
++ *
++ * Copyright (c) 2026 The NextBSD Project
++ *
++ * Redistribution and use in source and binary forms, with or without
++ * modification, are permitted provided that the following conditions
++ * are met:
++ * 1. Redistributions of source code must retain the above copyright
++ *    notice, this list of conditions and the following disclaimer.
++ * 2. Redistributions in binary form must reproduce the above copyright
++ *    notice, this list of conditions and the following disclaimer in the
++ *    documentation and/or other materials provided with the distribution.
++ *
++ * THIS SOFTWARE IS PROVIDED BY THE AUTHOR AND CONTRIBUTORS ``AS IS'' AND
++ * ANY EXPRESS OR IMPLIED WARRANTIES ARE DISCLAIMED.
++ */
++
++/*
++ * Peripheral clock gates for RP1, the southbridge on Raspberry Pi 5-family
++ * boards.
++ *
++ * Deliberately not a clock tree, and the scope is the whole design decision
++ * here. RP1's clockman drives three PLLs, a dozen derived sources and around
++ * forty leaf clocks; the vendor driver for it is 2500 lines. None of that is
++ * needed to make one peripheral tick. The leaf clocks in the table below take
++ * a fractional divider straight off the 50MHz crystal, and that crystal is
++ * already running -- everything on RP1 that works today works because it is.
++ *
++ * So this exposes one operation, "run clock N at F Hz", for the clocks it
++ * names, and returns ENOENT for everything else rather than guessing. A clock
++ * whose parent is a PLL is not in the table and must not be added to it
++ * without also bringing that PLL up, which is exactly the part this file does
++ * not do. Silently divide-by-zeroing off a dead PLL would look like working
++ * code and produce a dead peripheral.
++ *
++ * Register offsets and the divider arithmetic are from the vendor clk-rp1.c.
++ * Parent selection is the one part worth spelling out: for these clocks the
++ * CTRL register's SRC field is unused -- the vendor's table gives them
++ * clk_src_mask = 0, so its set_parent writes nothing there -- and selection is
++ * entirely by AUXSRC, which indexes that clock's own auxiliary parent list.
++ * xosc sits at index 2 in both PWM lists.
++ *
++ * UNPROVEN ON HARDWARE. Written against the vendor driver and the device tree,
++ * not against a board: the only Pi 5-family machine available has no fan, and
++ * its DTB leaves both PWM blocks disabled. Read the attach output before
++ * trusting it.
++ */
++
++#include <sys/param.h>
++#include <sys/systm.h>
++#include <sys/bus.h>
++#include <sys/kernel.h>
++#include <sys/lock.h>
++#include <sys/module.h>
++#include <sys/mutex.h>
++#include <sys/rman.h>
++
++#include <machine/bus.h>
++
++#include <dev/ofw/ofw_bus.h>
++#include <dev/ofw/ofw_bus_subr.h>
++
++#include <arm/broadcom/bcm2835/rp1_clk.h>
++
++/* CTRL, common to every leaf clock. */
++#define	RP1_CLK_CTRL_ENABLE		(1u << 11)
++#define	RP1_CLK_CTRL_AUXSRC_MASK	0x000003e0u
++#define	RP1_CLK_CTRL_AUXSRC_SHIFT	5
++
++/* The divider is 16.16 fixed point, split across two registers. */
++#define	RP1_CLK_DIV_FRAC_BITS		16
++#define	RP1_CLK_DIV_ONE			(1ULL << RP1_CLK_DIV_FRAC_BITS)
++#define	RP1_CLK_DIV_INT_MAX		0xffffULL
++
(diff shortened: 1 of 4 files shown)


pkgdemon:
Moved to draft to keep the focus on #171.

Not because anything here is known-broken — CI is green and the series applies — but because #171 turned out to carry a **fix to the BCM2712 mailbox DMA clamp** that silently kills the framebuffer console, and that wants testers' full attention on one kernel first.

Two concrete reasons this should not be reviewed in parallel:

1. **It is stacked on #171's branch**, which has now been restructured (the vcbus fix inserted first, the firmware patches renumbered 0030/0031 → 0031/0032). This branch needs a corresponding renumber to 0032/0033/0034 before it is coherent again — I will do that once #171 settles, rather than churning it while #171 is still moving.

2. **It cannot be tested yet regardless.** `rp1_pwm` needs a Pi 5 B with an Active Cooler fitted; it will not even attach on a Pi 500+, where both `cooling_fan` and `rp1_pwm1` are `status = "disabled"`. Nothing in the current testing round exercises it.

Nothing to re-review when it comes back — the patch contents are unchanged and the renumber only touches filenames and `patches/series`.

pkgdemon:
## How to test this kernel — the fan, and only the fan

You are replacing **one file** — `kernel8.img` on the FAT32 boot partition. Nothing else changes, and the old kernel stays on the partition so you can put it back.

### What is in this build, and what deliberately is not

Worth reading before anything else, because it is the reason to use *this* artifact and not a fresher one.

It is built from `73d967b2` on `feat/112-rp1-pwm-fan` — **#171 exactly as it merged, plus these three patches and nothing else**:

| | |
|---|---|
| `0001`–`0034` | everything that landed as #171, including the mailbox DMA clamp and the temperature sysctls |
| `0035` | `rp1_clk` — RP1 peripheral clock gates |
| `0036` | `rp1_gpio` — `fdt_pinctrl` provider, the pin mux |
| `0037` | `rp1_pwm` — the PWM driver |

**Not in it**, on purpose:

- the LinuxKPI/DRM groundwork now on `main` (`0038`–`0041`, #176 — `dma_alloc_wc`, `dma_mmap_wc`, the platform-device and component shims)
- the three zombie fixes that landed after #171 (#177 / #179 / #180)

This branch has not been rebased onto current `main`, and for this test that is a feature rather than staleness: nothing in the KMS work in progress can confound a fan result. If the fan does not spin, `rp1_clk`, `rp1_gpio` and `rp1_pwm` are the only new things it could be.

The corollary — **this run is the isolated artifact.** Once the branch is rebased the DRM patches come with it and a rebuild stops being a clean test of the fan alone. Please don't substitute a newer build.

### What you need

A **Pi 5 B (or a CM5 on a carrier) with the Active Cooler or the official case fan fitted** to the 4-pin JST header.

A Pi 500 or 500+ cannot test this at all — no fan header, and its DTB leaves both `cooling_fan` and `rp1_pwm1` at `status = "disabled"`, so `rp1_pwm` will not even attach. If that is your board, there is nothing new here for you.

### 1. Get the kernel

Download the **`rpi5-kernel8-arm64`** artifact from:

**https://github.com/nextbsd/nextbsd-kernel/actions/runs/33755765173**

Scroll to **Artifacts** at the bottom of that page. You need to be signed in to GitHub to download it — that is a GitHub restriction on Actions artifacts, not a choice here. It expires **2026-12-02**.

Or with the `gh` CLI:

```sh
gh run download 33755765173 --repo nextbsd/nextbsd-kernel -n rpi5-kernel8-arm64
```

Either way you get a zip containing a single file, `kernel8.img` (~15 MB). Unzip it.

Sanity-check you have the right one before copying it anywhere:

```sh
strings -a kernel8.img | grep -m1 "NextBSD Kernel Version"
# expect: NextBSD Kernel Version 20260903-123738
```

### 2a. If you boot from the USB stick

Plug the stick into any computer. The first partition is FAT32, labelled **NEXTBSD**, and every desktop OS mounts it automatically — macOS at `/Volumes/NEXTBSD`, Linux usually under `/media/…`, Windows as a drive letter.

```sh
cd /Volumes/NEXTBSD          # or wherever it mounted

cp kernel8.img kernel8.bak   # keep the working one — this is your way back
cp /path/to/new/kernel8.img kernel8.img
```

Then **eject properly** rather than just unplugging — the copy sits in the write cache otherwise and you will boot a truncated kernel. `Eject` in Finder, the eject icon on Linux, "Safely Remove" on Windows. Or from a terminal: `sync && diskutil unmount /Volumes/NEXTBSD` (macOS) / `sync && umount /media/…` (Linux).

### 2b. If you already installed to NVMe

The boot partition is on the NVMe, not the stick, so do it from the running system — SSH in or use the console:

```sh
mkdir -p /mnt/efi
mount -t msdosfs /dev/nda0s1 /mnt/efi        # check `gpart show` if nda0 is not your disk

cp -p /mnt/efi/kernel8.img /mnt/efi/kernel8.bak
cat /path/to/new/kernel8.img > /mnt/efi/kernel8.img
chmod 755 /mnt/efi/kernel8.img

sync; umount /mnt/efi
```

### 3. Turn on verbose boot — this build wants it

`0036` logs the FUNCSEL it found on the pin *before* overwriting it, at `bootverbose`. That line answers a question about hardware nobody here can answer: whether the Pi 5 B firmware has already muxed GPIO45 before we get there. Without verbose boot the rest of the test still works, but that data point is lost.

There is no loader on this boot path — no `loader.conf`, no `boot -v` prompt. Boot flags come from **`cmdline.txt`** on the same FAT32 partition, behind a `FreeBSD:` guard. Everything ahead of the guard is the firmware's Linux vocabulary; everything after it is ours. Append to the end of the existing single line:

```
 FreeBSD: -v
```

`cmdline.txt` is one line — add it to the end of that line, don't start a new one.

This only works because `0021` is in this kernel: the firmware prepends a dozen of its own parameters before whatever `cmdline.txt` holds, and the stock check requires the guard at position 0, so it never matched on a Pi and discarded every flag silently. On an older kernel the line above does nothing and looks identical to it working.

### 4. Boot, and check the three drivers attached

```sh
uname -v                                              # expect 20260903-123738
dmesg | grep -iE 'rp1_clk|pwm|funcsel'
ls /dev/pwm/
```

What you want to see:

```
rp1_clk0: <RP1 peripheral clocks> ...
pwm0: <RP1 PWM> ...
pwm0: clock 50000000 Hz
gpio0: gpio45 funcsel 31 -> 0 (pwm1)
```

and `/dev/pwm/` holding `pwmc0.0` … `pwmc0.3`.

Note the PWM controller registers under the name `pwm`, so it is `pwm0`, not `rp1_pwm0`.

That funcsel line settles the open question either way:

```
gpio0: gpio45 funcsel 31 -> 0 (pwm1)     # 31 = NULL, firmware left it alone
gpio0: gpio45 funcsel 0 -> 0 (pwm1)      # firmware had already done it
```

Failures here print rather than going quiet, which was the point of writing them that way:

| message | means |
|---|---|
| `pwm0: could not resolve the clocks property` | the DT lookup failed |
| `pwm0: could not start the PWM clock: N` | `rp1_clk` refused it |
| `pwm0: PWM clock reads back as stopped` | gate written, nothing ticking |
| `pwm0: pinctrl "default" not applied: N` | the mux failed (bootverbose only) |
| no `pwm0` at all | the block is `disabled` in your DTB — wrong board |

### 5. Drive the fan by hand

The fan is channel 3 of the first controller. Use whatever `ls /dev/pwm/` actually showed rather than assuming — `pwm -f` takes the name relative to `/dev/pwm/`.

```sh
pwm -f pwmc0.3 -C                    # what the driver thinks it is doing
pwm -f pwmc0.3 -p 41566 -d 50% -E    # ~50% — should be audible
pwm -f pwmc0.3 -p 41566 -d 100% -E   # full
pwm -f pwmc0.3 -D                    # off
```

`41566` is the period the device tree asks for — `pwms = <&rp1_pwm1 3 41566 PWM_POLARITY_INVERTED>`, nanoseconds, about 24 kHz.

**If the fan runs backwards** — full blast when you asked for off, silent at 100% — that inverted polarity is why. Add `-I`:

```sh
pwm -f pwmc0.3 -p 41566 -d 50% -I -E
```

⚠️ **`-I` is not sticky.** `pwm(8)` *clears* the inverted flag on every invocation that does not pass it, so once you know you need it, put `-I` on every command. That is `pwm(8)` behaviour in base, not something this driver does.

### 6. What to report

1. **Does it spin?** At 50% and at 100% — and did you need `-I`?
2. **The funcsel line from `dmesg`, verbatim.** This is the one thing that cannot be determined without your board.
3. **`pwm -f pwmc0.3 -C`** output after setting a duty cycle.
4. **Which board and which fan** — Pi 5 B or CM5, Active Cooler or official case fan.
5. If nothing attached at all: `dmesg | grep -iE 'rp1|pwm'` in full.

A negative result is as useful as a positive one. "attached, `/dev/pwm/pwmc0.3` exists, no air moving" and "`pwm0` never attached" are different bugs and both are worth having.

One thing that is **not** a bug: the fan will not step itself with temperature. That policy is deliberately not in this PR (see "What's still missing" in the description). Nothing here changes fan behaviour unless you drive it by hand, so an untouched boot on this kernel should look exactly like the #171 build did.

### Going back

```sh
cp kernel8.bak kernel8.img     # from wherever you made the backup
```

Take the ` FreeBSD: -v` back out of `cmdline.txt` if you want the quieter boot. Eject or `sync; umount` again, and reboot.

probonopd:
Let me know if it would be helpful for you to have access to a RPi 5 with the original active fan over ssh. I can set that up for you if it helps.

pkgdemon:
> Let me know if it would be helpful for you to have access to a RPi 5 with the original active fan over ssh. I can set that up for you if it helps.

That might be useful yes.  I think we should probably just merge this and try since the kernel will be often mismatched from the latest image unless I rebase it constantly while I am working on KMS for videocore.

pkgdemon:
Closing this PR since it was merged in https://github.com/nextbsd/nextbsd-kernel/pull/191 and will be part of next image rebuilds after videocore works...

PR #174, opened 2026-09-02
kern_exit: reap orphans adopted by init, the way XNU does (#173)

### LINKED ISSUE #173: Orphans reparented to PID 1 are never reaped: zombies accumulate without bound
## Symptom

`<defunct>` processes accumulate without bound on NextBSD. Reported by @probonopd on
gershwin-on-nextbsd; reproducible on demand.

```
$ ps -axo state= | grep -c Z
4                       # after boot
... run two throwaway orphans ...
13
```

They are never collected. Neither FreeBSD nor Linux nor macOS shows this with the same
Gershwin userland, which is what makes it ours.

## Root cause

Orphans reparented to PID 1 are never reaped, because on NextBSD PID 1 is launchd and
launchd has no steady-state stray reaper.

Measured on the box (kernel `20260901-032651`, arm64):

```
A (plain orphan, never talks to launchd)   -> Z   not reaped
B (orphan that execs launchctl)            -> Z   not reaped
```

The zombies are the short-lived halves of double-fork daemonisations — `gdnc`, `gpbs`,
`dbus-launch`. Each pairs with a live daemon:

```
176    1 Ss   gdnc          <- live
288    1 Z    gdnc          <- ZOMBIE
290    1 Z    gdnc          <- ZOMBIE
278    1 Is   dbus-daemon   <- live
277    1 Z    dbus-launch   <- ZOMBIE
```

The spawner forks, the stub forks the real daemon and exits, and nobody calls `wait()`.
That is a real bug in Gershwin/GNUstep and worth fixing there too — but it exists
identically on FreeBSD and Linux, where it is harmless because PID 1 cleans up. The
accumulation is ours.

Rate is proportional to session and app launches, not time: every one of the four
boot-time zombies was created inside a 10-second window at session startup, and an idle
box stays flat. That is why the reproduction is "open and close Terminal repeatedly".

## Why the other three systems are immune

| PID 1 | reaps strays? | how |
|---|---|---|
| macOS launchd | yes | XNU reaps at exit — measured: orphan goes `S -> gone`, no `Z` window at 20ms polling |
| FreeBSD `init(8)` | yes | `waitpid(-1, …, WNOHANG)` — init.c:971, 1136, 1992, plus blocking waits at 1749, 1922 |
| Linux systemd | yes | reaps as PID 1 |
| **NextBSD launchd** | **no** | nothing |

Apple's is **not** `SA_NOCLDWAIT` in launchd: `launchctl list` on macOS still reports
`LastExitStatus` (`com.apple.knowledgeconstructiond` shows `-9`), which it could not if
launchd had that flag set. The only arrangement consistent with both observations is
kernel-side reaping of orphans, with launchd's own job bookkeeping untouched.

## Fix: give the kernel a reap-at-exit path for init's orphans

Match XNU rather than papering over it in launchd.

### The primitive already exists

`procdesc_close()` (`sys/kern/sys_procdesc.c:386`) already reaps a zombie with no waiter
at all:

```c
sx_xlock(&proctree_lock);
PROC_LOCK(p);
if (p->p_state == PRS_ZOMBIE)
        proc_reap(curthread, p, NULL, 0);
```

So `proc_reap()` is callable outside the wait paths. Requirements: `proctree_lock`
XLOCKED, `PROC_LOCK(p)`, `PRS_ZOMBIE`, and any thread as `td` (it charges rusage to
`td->td_proc`). It consumes both locks on every path, so a sweep must re-scan rather than
iterate a list.

### The correctness constraint that shapes the design

**A blanket "reap orphans of initproc" would be wrong.** launchd's own jobs are children
of PID 1 too. Reaping those steals the exit status from `job_reap()`'s `wait4()`
(`core.c:3770`), which silently breaks `KeepAlive`, `LastExitStatus` and crash-restart —
the same defect as `SA_NOCLDWAIT`, relocated into the kernel where it is harder to see.

The distinction is available for free. `exit1()` at kern_exit.c:506 is where a dying
parent hands its untraced children to the reaper, and nothing else goes through it:

```c
if ((q->p_flag & P_TRACED) == 0) {
        proc_reparent(q, q->p_reaper, true);
```

So flag there. Then:

- forked by launchd, never reparented -> no flag -> untouched, `wait4()` works
- parent died, reparented to init -> flagged -> auto-reaped, never a visible zombie

That is exactly the `gdnc`/`gpbs`/`dbus-launch` stubs and nothing else.

### Shape

1. `sys/sys/proc.h` — new `P2_ORPHAN_AUTOREAP` bit (`0x04000000`; highest in use today is
   `P2_HWT 0x02000000`).
2. `kern_exit.c` orphan reparent site — set the flag when reparenting an untraced child to
   the reaper.
3. `kern_exit.c` exit path — when such a process exits with `p_pptr == initproc`, schedule
   a deferred sweep. The dying thread cannot reap itself: it is not `PRS_ZOMBIE` until
   line 714, `PROC_SLOCK` is held from there to `thread_exit()`, and `td->td_proc` is the
   corpse.
4. Taskqueue handler — sweep initproc's children for flagged zombies and reap each. Sweeps
   by scan rather than by pid, so there are no dangling `struct proc` pointers and the
   work coalesces. Re-arms if it sees a flagged child that has not reached `PRS_ZOMBIE`
   yet, so correctness does not depend on timing.

Gated on `p_pptr == initproc` specifically, so a userland reaper established with
`procctl(PROC_REAP_ACQUIRE)` still gets its children and can wait for them normally.

## Alternatives considered

- **`SA_NOCLDWAIT` in launchd** (the "recommended" option in the original writeup) —
  rejected. Auto-reaps every child of launchd including tracked jobs, destroying the exit
  statuses `job_reap()` depends on. Also not what Apple does.
- **Periodic `waitpid(-1, WNOHANG)` sweep in launchd** — this is FreeBSD's `init`
  approach, not Apple's. Workable, and a reasonable interim, but it must peek with
  `waitid(..., WNOWAIT)` and skip pids that `jobmgr_find_by_pid()` knows, or it races
  `job_reap()` and steals statuses intermittently. Leaves a zombie window of one sweep
  interval.

## Not covered by this fix

`gpbs` pid 200 has PPID 23 (LoginWindow), which is **alive** and simply never calls
`wait()`. No PID 1 fix can reach that one — the parent has to reap it. Previously fixed
once in Gershwin's LoginWindow (2026-03-23) and evidently regressed or never covered this
path. Tracked separately, Gershwin-side.

## Notes

Two things ruled out during investigation, recorded so they are not re-chased:

- **Not the `launchctl list` filter.** `9230aca` hides only the synthetic bootstrapper
  from the listing; jobs and reaping are untouched.
- **Not epic #145.** Port names colliding with the fd namespace produce wedges, not
  unreaped children.

One genuinely open and unrelated question surfaced: the box shows **zero anonymous jobs**
(`launchctl list | grep -c anonymous` -> 0) on a running desktop, which should not be
possible for launchd-842. That is worth its own issue — but it is not the cause here,
since the stubs exit immediately and never make a Mach call, so they could never have had
an anonymous job even on a healthy launchd.

Full investigation writeup:
https://gist.github.com/pkgdemon/6385dec92702ae298e001e67511ad67d

pkgdemon:
Fixes #173.

## The bug

PID 1 on NextBSD is launchd, and launchd has no steady-state `waitpid(-1)` loop the way
`init(8)` does — Apple's doesn't either. An orphan reparented onto PID 1 therefore stays a
zombie **forever**. Measured on gershwin-on-nextbsd:

```
A (plain orphan, never talks to launchd)   -> Z   not reaped
B (orphan that execs launchctl)            -> Z   not reaped
total zombies                              4 -> 13
```

Every orphan leaks. A desktop that double-forks its helpers (`gdnc`, `gpbs`,
`dbus-launch`) accumulates them without bound.

## Why the other three systems are immune

| PID 1 | reaps strays? | how |
|---|---|---|
| macOS launchd | yes | XNU reaps at exit — measured on 26.6.2, orphan goes `S → gone`, **no `Z` window** at 20ms polling |
| FreeBSD `init(8)` | yes | `waitpid(-1, …, WNOHANG)` — init.c:971, 1136, 1992, +1749, 1922 |
| Linux systemd | yes | reaps as PID 1 |
| **NextBSD launchd** | **no** | nothing |

Apple's is **not** `SA_NOCLDWAIT` in launchd — `launchctl list` on macOS still reports
`LastExitStatus` (`com.apple.knowledgeconstructiond` shows `-9`), which it couldn't if that
flag were set. The only arrangement consistent with both is kernel-side reaping with
launchd's own bookkeeping untouched. That's what this does.

## The safety property that shapes the whole design

> reap **only** processes orphaned onto the reaper, never processes PID 1 forked itself.

launchd's own jobs are children of PID 1 too, and `job_reap()` collects their status with
`wait4()` (`core.c:3770`). Reaping those out from under it would silently break
`KeepAlive`, `LastExitStatus` and crash-restart — the same defect as `SA_NOCLDWAIT`,
relocated into the kernel where it's harder to find.

`P2_ORPHAN_AUTOREAP` is set at exactly one place — the orphan-reparent site in `exit1()`,
which nothing but a dying parent's untraced children reaches. The two populations never
mix.

## Three constraints from the existing exit path

Not design choices — properties of the code:

- **The reap can't run on the exiting thread.** The process isn't `PRS_ZOMBIE` until
  kern_exit.c:714, `PROC_SLOCK` is held from there through `thread_exit()`, and
  `proc_reap()` wants a `td` whose `td_proc` is somebody else. Hence the taskqueue.
- **The sweep scans rather than carrying a pid.** A pid would be a dangling `struct proc`
  the moment something else reaped it first; a scan can't be, and concurrent exits coalesce
  into one pass. `proc_reap()` consumes `proctree_lock` *and* the proc lock on every path,
  so each iteration re-scans rather than walking a list.
- **The orphan's parent is deliberately not signalled.** Leaving `signal_parent` at 0
  avoids queueing a SIGCHLD nothing consumes and — load-bearing — keeps `p_ksi` off init's
  sigqueue. `proc_reap()` calls `sigqueue_take()` under `PROC_LOCK(td->td_proc)`, and our
  `td` belongs to the taskqueue thread, not init, so an enqueued ksi would be manipulated
  under the wrong lock. This one bit me in the first draft.

## Not a new primitive

`procdesc_close()` (`sys_procdesc.c:386`) already does this:

```c
sx_xlock(&proctree_lock);
PROC_LOCK(p);
if (p->p_state == PRS_ZOMBIE)
        proc_reap(curthread, p, NULL, 0);
```

Zombie, no waiter, outside any wait path, parent likewise never signalled. This is the
same shape applied to a second case.

## Scope

Gated on `p_pptr == initproc`, so a userland reaper established with
`procctl(PROC_REAP_ACQUIRE)` still collects its own children normally.

**Not covered, and not coverable here:** a zombie whose parent is *alive* and simply never
calls `wait()`. That's a bug in the parent — on this system, `gpbs` under LoginWindow
(pid 200, PPID 23). Gershwin-side, tracked separately.

## Testing

- Full series applies cleanly to `releng/15.1` at `88e7371d9dc`.
- CI builds `NEXTBSD` (amd64 + arm64) and `NEXTBSD-RPI5`.

**Not yet proven on hardware.** The check is direct — on a booted image:

```sh
ps -axo state= | grep -c Z                       # baseline
sh -c '( sleep 1; exit 0 ) & exit 0'             # make an orphan
ps -axo state= | grep -c Z                       # must not have grown
launchctl list | head                            # jobs still tracked, PIDs live
```

The second thing to confirm is that launchd's own accounting is intact — stop and start a
job and check `LastExitStatus` still updates. That's the property `SA_NOCLDWAIT` would
have destroyed, and the reason this patch is shaped the way it is.

### DIFF
--- a/patches/0029-kern_exit-reap-orphans-adopted-by-init-like-xnu.patch
+++ b/patches/0029-kern_exit-reap-orphans-adopted-by-init-like-xnu.patch
@@ -1,0 +1,268 @@
+From 4ad6ee76d2adbc42d55a4b2919e632c0f340108a Mon Sep 17 00:00:00 2001
+From: pkgdemon <jpm820@gmail.com>
+Date: Wed, 2 Sep 2026 17:30:30 -0500
+Subject: [PATCH] kern_exit: reap orphans adopted by init, the way XNU does
+
+PID 1 on NextBSD is launchd, and launchd has no steady-state waitpid(-1)
+loop the way init(8) does -- Apple's does not either. An orphan reparented
+onto PID 1 therefore stays a zombie forever, and a desktop session that
+double-forks its helpers accumulates them without bound. Measured on
+gershwin-on-nextbsd: every orphan leaks, and the count only ever grows.
+
+FreeBSD does not hit this because init(8) sweeps (init.c:971, 1136, 1992,
+plus blocking waits at 1749 and 1922). Linux does not because systemd
+does. macOS does not because XNU reaps such a process in the kernel at
+exit -- measured on macOS 26.6.2, an orphan goes S -> gone with no zombie
+window at all, polling every 20ms.
+
+Apple's is NOT SA_NOCLDWAIT in launchd: `launchctl list` on macOS still
+reports LastExitStatus, which it could not if that flag were set. The only
+arrangement consistent with both observations is kernel-side reaping of
+orphans with launchd's own bookkeeping left alone. This does that.
+
+The distinction is the whole design, because it is what makes it safe:
+
+	reap ONLY processes orphaned onto the reaper, never processes
+	that PID 1 forked itself.
+
+launchd's own jobs are children of PID 1 too, and job_reap() collects
+their status with wait4(). Reaping those out from under it would silently
+break KeepAlive, LastExitStatus and crash-restart -- the same defect as
+SA_NOCLDWAIT, relocated into the kernel where it is harder to find.
+P2_ORPHAN_AUTOREAP is set at exactly one place, the orphan-reparent site
+in exit1(), which nothing but a dying parent's untraced children reaches.
+The two populations never mix.
+
+Three things this has to work around, all of them properties of the
+existing exit path rather than choices:
+
+The reap cannot run on the exiting thread. The process is not PRS_ZOMBIE
+until late in exit1(), PROC_SLOCK is held from there through
+thread_exit(), and proc_reap() wants a td whose td_proc is somebody else.
+So it is deferred to the thread taskqueue.
+
+The sweep scans initproc's children rather than carrying a pid. A pid
+would be a dangling struct proc the moment something else reaped it first;
+scanning cannot be, and concurrent exits coalesce into one pass. proc_reap()
+consumes proctree_lock and the proc lock on every path, so each iteration
+re-scans rather than walking a list.
+
+The orphan's parent is deliberately not signalled. Leaving signal_parent
+at 0 avoids queueing a SIGCHLD nothing will consume, and -- load-bearing --
+keeps p_ksi off init's sigqueue: proc_reap() calls sigqueue_take() under
+PROC_LOCK(td->td_proc), and our td belongs to the taskqueue thread rather
+than init, so an enqueued ksi would be manipulated under the wrong lock.
+
+None of this is a new primitive. procdesc_close() (sys_procdesc.c:386)
+already calls proc_reap(curthread, p, NULL, 0) on a zombie with no waiter,
+from outside any wait path and with the parent likewise never signalled.
+This is the same shape applied to a second case.
+
+Gated on p_pptr == initproc specifically, so a userland reaper established
+with procctl(PROC_REAP_ACQUIRE) still collects its own children normally.
+
+Not covered, and not coverable here: a zombie whose parent is alive and
+simply never calls wait(). That is a bug in the parent; on this system the
+example is gpbs under LoginWindow.
+
+Refs nextbsd-kernel#173.
+---
+ sys/kern/kern_exit.c | 125 ++++++++++++++++++++++++++++++++++++++++++-
+ sys/sys/proc.h       |   3 ++
+ 2 files changed, 127 insertions(+), 1 deletion(-)
+
+diff --git a/sys/kern/kern_exit.c b/sys/kern/kern_exit.c
+index 667545bd6ac..fad3c2c752f 100644
+--- a/sys/kern/kern_exit.c
++++ b/sys/kern/kern_exit.c
+@@ -51,6 +51,7 @@
+ #include <sys/malloc.h>
+ #include <sys/mutex.h>
+ #include <sys/proc.h>
++#include <sys/taskqueue.h>
+ #include <sys/procdesc.h>
+ #include <sys/ptrace.h>
+ #include <sys/racct.h>
+@@ -221,6 +222,89 @@ proc_set_p2_wexit(struct proc *p)
+  * zombie, and unlink proc from allproc and parent's lists.  Save exit status
+  * and rusage for wait().  Check for child processes and orphan them.
+  */
++/*
++ * Reap-at-exit for orphans adopted by init (nextbsd#173).
++ *
++ * PID 1 on NextBSD is launchd, which -- like Apple's -- has no steady-state
++ * waitpid(-1) loop the way init(8) does. An orphan reparented onto it therefore
++ * stays a zombie forever, and a desktop that double-forks its helpers piles them
++ * up without bound.
++ *
++ * XNU reaps such a process in the kernel at exit, which is why macOS never shows
++ * a zombie even for an instant. Do the same, with one distinction that is the
++ * entire reason this is safe:
++ *
++ *	reap ONLY processes that were orphaned onto the reaper, never
++ *	processes that PID 1 forked itself.
++ *
++ * launchd's own jobs are children of PID 1 too, and it collects their status
++ * with wait4(). Reaping those out from under it would silently break
++ * KeepAlive/LastExitStatus -- the same defect as setting SA_NOCLDWAIT in
++ * launchd, only harder to find. P2_ORPHAN_AUTOREAP is set at exactly one place,
++ * the orphan-reparent site in exit1(), so the two populations never mix.
++ *
++ * The reap cannot run on the exiting thread: the process is not PRS_ZOMBIE until
++ * late in exit1(), PROC_SLOCK is held from there through thread_exit(), and
++ * proc_reap() wants a td whose td_proc is somebody else. Hence the taskqueue.
++ *
++ * The sweep scans rather than carrying a pid, so it can never be left holding a
++ * freed struct proc if something reaps the process first, and concurrent exits
++ * coalesce into one pass. proc_reap() consumes proctree_lock and the proc lock
++ * on every path, hence a re-scan per iteration rather than a list walk.
++ */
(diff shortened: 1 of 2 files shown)


pkgdemon:
## Verified on hardware

Deployed to `gershwin-on-nextbsd` (arm64) via `nextboot` and rebooted. Running
`NextBSD Kernel Version 20260902-223544` from this PR's CI build.

### The fix works

| | before patch | after patch |
|---|---|---|
| zombies after boot | **4** | **1** |
| after creating 5 orphans | would be 9 | **still 1** |
| launchd jobs tracked | 17 | 17 |

The three PPID=1 orphans that boot used to leave behind — `dbus-launch`, `gdnc`, `gdnc` —
are gone, reaped by the kernel. Five deliberately created orphans left zero residue; on
the old kernel every single one leaked.

### Both negative controls behave correctly

This is what says the patch isn't over-reaching:

- The one remaining zombie is `200 23 Z gpbs` — PPID 23 is **LoginWindow, which is
  alive**. Parent-alive-never-waits is exactly the case this patch deliberately does not
  touch, and it is untouched.
- A synthetic control (live parent that never `wait()`s) **stayed a zombie** through the
  test. If the flag were being set too broadly, that would have disappeared too.

### The safety property, tested directly

The whole design rests on launchd still collecting its own jobs' exit statuses. Killed a
tracked job with SIGTERM:

```
$ launchctl list com.apple.aslmanager      # before
        "LastExitStatus" = 0;
        "PID" = 48;

$ kill -TERM 48

$ launchctl list com.apple.aslmanager      # after
        "LastExitStatus" = 15;             <- SIGTERM, collected correctly
$ ps -p 48
pid 48 gone (reaped by launchd, no zombie)
```

`LastExitStatus` updated to 15. The kernel did **not** steal the status from
`job_reap()`'s `wait4()` — which is precisely what `SA_NOCLDWAIT` would have done, and the
reason this patch distinguishes orphans from PID 1's own children. Job table still 17
jobs, all tracked with live PIDs.

(`com.apple.aslmanager` shows as not running afterwards, as expected — it is
demand-launched and returns on the next syslogd trigger.)

### Still outstanding

Not yet exercised: a full desktop session with repeated Terminal open/close, which is
@probonopd's original reproduction and the highest-rate case. The kernel-side behaviour is
proven; that would confirm the end-user symptom is gone.

pkgdemon:
Correcting one number in my verification comment above, for the record.

I quoted "zombies after boot: 4 before, 1 after" as evidence. **That comparison was not
sound** and should be disregarded. The two boots were not equivalent: the 4-zombie boot had
two GNUstep session waves, the 1-zombie boot had one. Boot-time zombie count tracks how
many sessions started, not whether the patch works.

The valid evidence is the synthetic orphan test, which I have since run on both kernels on
the same box in the same state:

| identical test | old kernel `20260901-032651` | patched `20260902-223544` |
|---|---|---|
| zombies before | 1 | 1 |
| after 5 orphans | **6** | **1** |
| the 5 orphans | all `Z`, PPID 1 | all reaped |

```
# old kernel
zombies before: 1
zombies after 5 orphans: 6
207   23 Z    gpbs      <- pre-existing, parent alive, out of scope
268    1 Z    sh
271    1 Z    sh
274    1 Z    sh
277    1 Z    sh
280    1 Z    sh
```

Five orphans in, five zombies out on the old kernel; zero on the patched one. That is not
confounded by session count, and it is the result the patch should be judged on.

The other two findings in the comment above stand unchanged and were never affected by
this: the live-parent control still becomes a zombie (so the patch is not over-reaping),
and `LastExitStatus` still updated to 15 after a tracked job was SIGTERMed (so launchd's
own accounting is intact).

Issue #51, opened 2026-07-14
net80211: wlan_xauth missing — every WPA/WPA2/WPA3 association fails EINVAL


pkgdemon:
## Symptom

Every WPA / WPA2 / WPA3 network fails to associate. Open networks work. `wpa_supplicant` never leaves `SCANNING` and surfaces no useful error.

## Root cause

`ioctl(SIOCS80211, IEEE80211_IOC_AUTHMODE)` returns `EINVAL`, always.

`wpa_supplicant` sets the authmode to `IEEE80211_AUTH_WPA`. `ieee80211_ioctl.c:2858-2870` calls `ieee80211_authenticator_get()`, which (`ieee80211_proto.c:471-478`) looks up `auth_modnames[IEEE80211_AUTH_WPA]` = `"wlan_xauth"` and tries `ieee80211_load_module("wlan_xauth")` → `kldload` → **no such module on disk** → returns NULL → `EINVAL`. Association never begins.

Two independent facts combine, and **neither is wrong on its own**:

1. **`wlan_xauth` is not in the kernel.** Our config is `include GENERIC` plus additions (`config/NEXTBSD:1`). Nothing wireless is excluded — `device wlan` + `wlan_ccmp`/`wlan_tkip`/`wlan_wep`/`wlan_gcmp`/`wlan_amrr` all arrive from GENERIC. But `wlan_xauth` is not in GENERIC either: upstream FreeBSD deliberately leaves it out, because net80211 autoloads it as a `.ko` on demand. On stock FreeBSD this is invisible and self-healing.

2. **NextBSD has no `.ko` tree to autoload from.** `ci/assemble-image.sh:168` does `rm -f "$ROOTFS"/boot/kernel/*.ko` — the whole module tree is deleted at image-assembly time; only `.kext` bundles survive. This is intentional and correct for the kext model.

Together they silently remove the recovery mechanism upstream depends on. **This is a regression NextBSD introduced, not an unimplemented feature.**

## Fix

```
# config/NEXTBSD
device      wlan_xauth   # 802.11 802.1x/WPA authenticator.
                         # net80211 autoloads this as a .ko on stock FreeBSD;
                         # NextBSD ships no .ko tree (assemble-image.sh:168),
                         # so it must be compiled in.
```

[...]

PR #199, opened 2026-09-06
linuxkpi: bind components in match order, and unbind with a real master

pkgdemon:
Two fixes in the component framework, both needed by the vc4 KMS port (nextbsd/nextbsd-kernel-extensions#51). Both were found by an adversarial review of the vc4 newbus shims rather than by a failing build — neither produces an error message.

## 1. Bind in match order

`component_bind_all()` walked the global component list, so components bound in the order they registered — for FDT drivers, device-tree order. Upstream Linux binds by indexing the match array instead (`drivers/base/component.c`, *"Bind components in match order"*).

The failure is silent. `vc4_crtc_bind()` calls `vc4_set_crtc_possible_masks()` (`vc4_crtc.c:1373-1397`), which walks the encoders **already registered** on the `drm_device` and sets `encoder->possible_crtcs`. Nothing revisits it. If a pixelvalve binds before the HDMI encoders exist, those encoders keep `possible_crtcs == 0` — DRM has no CRTC able to drive the HDMI connectors, fbdev finds no usable CRTC, and every modeset fails while the driver reports a successful load.

bcm2712's device tree is in exactly that order: pixelvalves at `0x7c410000` / `0x7c411000` come before the HDMI controllers at `0x7ef00700` / `0x7ef05700`, so registration order puts both CRTCs ahead of both encoders on every boot.

`component_unbind_all()` mirrors it in reverse. A master with no match list keeps the old registration-order walk — the single-component firmware-KMS case this framework was written for.

### What this is *not*

vc4's `component_drivers[]` is annotated *"The HDMI driver needs to be bound after the HVS so that we can lookup the HVS maximum core clock rate"*. That constraint is **stale** on `rpi-6.12.y`: the 4kp60 decision moved into `vc4_hvs_bind()` (`vc4_hvs.c:2125-2135`), which asks the firmware mailbox directly, and every remaining `vc4->hvs` use in `vc4_hdmi.c` is a runtime callback that cannot run before `drm_dev_register()`. Ordering matters here for `possible_crtcs`, not for the clock lookup.

## 2. Unbind with a real master

`component_del()` called `unbind(dev, NULL, NULL)`. Unbind callbacks read the master — `vc4_hvs_unbind()` (`vc4_hvs.c:2217-2220`) calls `dev_get_drvdata(master)` as its first statement — so `kldunload` of a bound pipeline was a NULL dereference rather than a teardown.

Components now record the master and master data they were bound with, and `component_del()` passes them.

## Testing

Full series applies clean; kernel builds on both arches. The runtime effect of (1) can't be observed until `vc4_kms.ko` links (nextbsd/nextbsd-kernel-extensions#51) — the ordering claim was verified against the vendored `bcm2712.dtsi` and a DTB read off the Pi 500+.

### DIFF
--- a/patches/0053-linuxkpi-rman-out-of-platform-device-h.patch
+++ b/patches/0053-linuxkpi-rman-out-of-platform-device-h.patch
@@ -1,0 +1,156 @@
+From 305bc01c830577f67e211309bbf0d5a12d9e5b6f Mon Sep 17 00:00:00 2001
+From: pkgdemon <jpm820@gmail.com>
+Date: Sat, 5 Sep 2026 21:39:36 -0500
+Subject: [PATCH] linuxkpi: keep sys/rman.h out of <linux/platform_device.h>
+
+Fixes a build regression from patch 0050, which added sys/rman.h to
+<linux/platform_device.h> so devm_platform_ioremap_resource() could be inline
+there.
+
+LinuxKPI declares its own struct resource in <linux/ioport.h> -- a different
+type, with different members, from FreeBSD's in sys/rman.h. Including
+sys/rman.h from platform_device.h put the FreeBSD definition ahead of the
+LinuxKPI one for every consumer of that header, so any driver that reached
+ioport.h afterwards failed with "redefinition of 'resource'". <linux/mfd/core.h>
+takes exactly that path -- it includes platform_device.h and then ioport.h --
+which broke the whole amdgpu build:
+
+  In file included from .../amdgpu.h:77:
+  In file included from .../amdgpu_acp.h:29:
+  In file included from .../linux/mfd/core.h:21:
+  .../linux/ioport.h:40:8: error: redefinition of 'resource'
+  .../sys/rman.h:102:8: note: previous definition is here
+
+The inline was unsound even where it compiled: "struct resource" inside it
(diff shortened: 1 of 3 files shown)


pkgdemon:
Closing — same reason as #202: this belongs in the module, not the kernel.

Both fixes here remain real and are not being abandoned:

1. **Bind in match order.** `component_bind_all()` binds in registration order; bcm2712's device tree lists both pixelvalves before both HDMI controllers, so `vc4_set_crtc_possible_masks()` would leave every encoder with `possible_crtcs == 0` and every modeset would fail silently.
2. **Unbind with a real master.** `component_del()` passed `NULL`, and `vc4_hvs_unbind()` dereferences the master immediately — a panic on kldunload, not a teardown.

But there is a further argument against landing this one in the kernel specifically: the match list it operates on is **always empty today**, because `platform_find_device_by_driver()` is deliberately unimplemented. The machinery would sit in a shared kernel header without yet being able to do its job.

Both fixes will be re-landed in `graphics/vc4_kms/lkpi/` in nextbsd/nextbsd-kernel-extensions, where the module is the only consumer and the drm-kmod build is the gate that actually catches regressions. Content preserved; branch left in place.

Refs nextbsd/nextbsd-kernel-extensions#51

PR #202, opened 2026-09-06
linuxkpi: real interrupt-names parsing, and more than one IRQ per device

pkgdemon:
Stacked on #199 (`fix/component-match-order`). Review that one first.

## What was wrong

`platform_get_irq_byname()` returned `dev->irq` for **any** name, and `request_irq()` allocated rid 0 for every device off the PCI bus. A device-tree device could therefore hold exactly one interrupt. The old comment was honest that this needed real parsing; this is that parsing.

bcm2712 needs more than one:

| block | interrupts |
|---|---|
| HVS | `ch0-eof`, `ch1-eof`, `ch2-eof` (`HVS_NUM_CHANNELS` = 3) |
| HDMI | `hpd-connected`, `hpd-removed`, `cec-rx`, `cec-tx` |

With a single rid the first request takes it and every later one fails. `vc4_hvs.c:2203` does **not** check the return, so it fails *silently*, leaving all three EOF handlers aliased to one line — miswired interrupts with no error message.

## How

`platform_get_irq()` and `platform_get_irq_byname()` return the resource **rid**, tagged with `LKPI_IRQ_OF`, and `request_irq()` allocates that rid rather than assuming 0. The tag bit sits far above any real interrupt number, so an untagged value still means "`dev->irq` at rid 0" — existing callers, firmware KMS included, are unaffected.

`platform_get_irq_byname()` resolves the name against the node's `interrupt-names`: NUL-separated strings, one per entry in `interrupts` and in the same order, so the index of the matching name *is* the rid. Both getters validate that index against the IRQ entries newbus already built from the node, so an out-of-range request returns `-ENXIO` at lookup instead of failing later as a resource allocation.

Both live in `linux_of.c` rather than the header — they read the device tree, and a LinuxKPI header that names FreeBSD resource types breaks its consumers (#200).

## Testing

Series applies clean. The multi-IRQ path has no runtime exercise yet — nothing in tree requests a second interrupt until `vc4_kms.ko` links (nextbsd/nextbsd-kernel-extensions#51). Single-IRQ behaviour is unchanged by construction and is what firmware KMS runs on hardware today.

### DIFF
--- a/patches/0055-linuxkpi-interrupt-names-multi-irq.patch
+++ b/patches/0055-linuxkpi-interrupt-names-multi-irq.patch
@@ -1,0 +1,270 @@
+From 3ce85301629c1d00f05827f35329d64a712f2ee0 Mon Sep 17 00:00:00 2001
+From: pkgdemon <jpm820@gmail.com>
+Date: Sat, 5 Sep 2026 21:47:20 -0500
+Subject: [PATCH] linuxkpi: real interrupt-names parsing, and more than one IRQ
+ per device
+
+platform_get_irq_byname() returned dev->irq for any name, and request_irq()
+allocated rid 0 for every device off the PCI bus, so a device-tree device could
+hold exactly one interrupt.
+
+bcm2712 needs more than one. The HVS requests three named interrupts
+(ch0-eof, ch1-eof, ch2-eof) and the HDMI controller four (hpd-connected,
+hpd-removed, cec-rx, cec-tx). With a single rid the first request takes it and
+every later one fails -- and vc4_hvs.c does not check the return, so it fails
+silently with all three handlers aliased to one interrupt line.
+
+platform_get_irq() and platform_get_irq_byname() now return the resource rid,
+tagged with LKPI_IRQ_OF, and request_irq() allocates that rid rather than
+assuming 0. The tag bit sits far above any real interrupt number, so an
+untagged value still means "dev->irq at rid 0" and existing callers -- firmware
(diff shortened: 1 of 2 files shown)


pkgdemon:
Closing — this belongs in the module, not the kernel.

The reasoning is the include order every drm-kmod module is built with:

```
-I linuxkpi/gplv2/include                      <- drm-kmod's own KPI      (first)
-I linuxkpi/bsd/include                        <- drm-kmod's own KPI
-I ${SYSDIR}/compat/linuxkpi/common/include    <- the FreeBSD kernel's    (third)
```

A header in the module's own directory shadows the kernel's for objects compiled into that module, and reaches nothing else — not `drm.ko`, not i915, not amdgpu, not the kernel. We already rely on this: drm-kmod patch `0003` extends `linuxkpi/gplv2/include/linux/dma-buf.h` rather than the kernel's copy.

11 of the 55 patches in `patches/` touch shared LinuxKPI headers, and both regressions we've had came from those 11 — the amdgpu build break (#200) and the `of_node` KBI boot loop. Adding more shared-header surface for a driver that does not yet link is the wrong trade.

The `interrupt-names` parsing and multi-IRQ work here is still needed and still correct; it will be re-landed in `graphics/vc4_kms/lkpi/` in nextbsd/nextbsd-kernel-extensions, where `vc4_kms.ko` is the only consumer. Content preserved; branch left in place.

Refs nextbsd/nextbsd-kernel-extensions#51

PR #208, opened 2026-09-11
Linuxulator on arm64, and CI that proves it on both arches

pkgdemon:
Closes the arm64 half of #60.

## Staged on purpose

This PR is pushed in two steps so the gate is demonstrated, not asserted:

1. **`1344c81` (pushed now)** — adds a build-time gate that the Linuxulator and FUSEFS are compiled in, on both matrix legs. **The arm64 leg is expected to go RED here.** That is the point: it makes the existing gap visible and proves the gate is not vacuous.
2. **the fix (pushed once (1) is red)** — wires the arm64 ABI and adds arm64 boot CI, turning it green.

## What was actually wrong

arm64 has had a complete Linux ABI in tree the whole time (`sys/arm64/linux/`). It was never wired because `src-overlay/conf/files.compat_linux` cannot express both arches — it names `amd64/linux/*` and pulls in three `x86/linux/*` files that do not exist on arm64. `sys/modules/linux64/Makefile` guards exactly those with `.if ${MACHINE_CPUARCH} == "amd64"`; `config(8)` has no equivalent, so the split has to be per-file.

Nothing caught it because an `optional compat_linux` entry that no option selects compiles nothing and leaves no trace — the arm64 leg built green while shipping no Linux ABI at all.

## FUSEFS

Already fine on arm64 — `options FUSEFS` is in `config/NEXTBSD` and `NEXTBSD-RPI5` does `include NEXTBSD`, and it is gated `optional fusefs` in the **MI** `sys/conf/files`. No change needed; the new gate just makes sure it stays that way.

## Also in here

- **arm64 boot CI.** Image assembly and upload were gated `matrix.target == amd64`, so the arm64 kernel was built and never booted. `ci/assemble-image.sh` already had an arm64 branch and all the arm64 `continuous` artifacts already exist — only the workflow guards were missing.
- **An arch bleed fix.** The kext download pattern `*kext*.tar.gz` matches `graphics-kexts-amd64.tar.gz` *and* `graphics-kexts-arm64.tar.gz`, so amd64 images were assembled with arm64 kexts laid in beside their own. Harmless there, actively wrong once arm64 assembles too.

## Known limits, stated plainly

- The arm64 boot job runs the **generic** arm64 kernel on `-M virt` under AAVMF. `NEXTBSD-RPI5` is not booted — qemu has no BCM2712 model, so Pi-specific breakage stays invisible to CI.
- This does not change the fact that smoke tests are `pull_request`-only and therefore merges to `main` still get no boot test. Widening that is a separate call.
- `smoke-test-arm64` is a first-run path. If it needs iteration on console or firmware details, that is expected.


pkgdemon:
Step 1 result — the gate works, and arm64 failed exactly where predicted.

`kernel (amd64, amd64)` → **success**, `kernel (arm64, aarch64)` → **failure**, at the new gate:

```
--- objects config(8) should have selected (arm64)
  MISSING linux_sysvec.o        MISSING linprocfs.o
  MISSING linux_machdep.o       MISSING linsysfs.o
  MISSING linux_misc.o          ok    fuse_vfsops.o
  MISSING linux_elf64.o         ok    fuse_vnops.o
--- symbols that must be in the linked kernel
  MISSING symbol elf_linux_sysvec
  ok    fuse_vfsops
```

Two things this pins down that were previously only inferred from reading config files:

1. **arm64 shipped no Linux ABI at all** — not a partial or misconfigured one. Every `compat_linux` object is absent and `elf_linux_sysvec` never linked.
2. **FUSEFS was already correct on arm64.** `fuse_vfsops.o`, `fuse_vnops.o` and the linked `fuse_vfsops` symbol are all present. No change was needed for FUSE; it just had nothing asserting it.

Pushing the fix commits now.

PR #210, opened 2026-09-11
patch: fix the linsysfs boot panic on multi-PCIe-domain arm64 (#209)

### LINKED ISSUE #209: Pi 500+ panics at boot: linsysfs PCI domain collision reaches pfs_purge() on an uninitialised vnode cache
## Summary

A Raspberry Pi 500+ panics ~1s into boot on the current `main` kernel. `LINSYSFS` was
compiled into arm64 for the first time by #208 (`0f9119d6`), and `config/NEXTBSD-RPI5`
does `include NEXTBSD`, so the Pi inherited it. `linsysfs_init()` collides two PCI device
names, and pseudofs's error path then dereferences NULL.

```
pfs_add_node: homonymous siblings: 'pci0000:00/0000:00:00.0' type 2
Fatal data abort:
 far: 0x0000000000000000   esr: 0x0000000096000004
  lr: 0xffff00000043ecec   elr: 0xffff000000440180
panic: vm_fault failed: 0xffff000000440180 error 1
Uptime: 1s
```

`esr 0x96000004` = EC 0x25 data abort, DFSC 0x04 (level-0 translation fault), WnR=0 — a
**read** of address 0.

## Root cause

1. `linsysfs.c:542` calls `linsysfs_run_bus(..., "/pci0000:00", "0000")`. The PCI domain is
   a hardcoded string literal; `linsysfs.c:290-294` formats every node as
   `"%s:%02x:%02x.%x"` from prefix/bus/slot/func. `dinfo->cfg.domain` exists
   (`pcivar.h:207`, filled at `pci.c:753`) and is even emitted by the uevent filler
   (`linsysfs.c:196-201`) — it is simply never used in the name.
2. The Pi 500+ has two PCIe host bridges. From this repo's own
   `patches/0013-bcm2838_pci-support-the-BCM2712-PCIe-controller.patch`: *"The Raspberry Pi
   500+ keeps everything useful behind PCIe — the NVMe on pcie1, and RP1 ... on pcie2."*
   `pci_host_generic_fdt.c` assigns `ecam = device_get_unit(dev)` as the PCI domain, so the
   two bridges are domains 0 and 1, each numbering from bus 0, each with a root port at
   00:00.0. Both render as `0000:00:00.0` under the single `/sys/devices/pci0000:00` node.
3. `pfs_add_node()` (`pseudofs.c:136`) prints the homonymous-siblings warning and returns
   `EEXIST`.
4. `pfs_create_dir()`'s failure path (`pseudofs.c:257-262`) calls `pfs_destroy(pn)` →
   `pfs_purge(pn)` (`pseudofs_vncache.c:300-326`).
5. **`pfs_purge()` walks `pfs_vncache_hashtbl[]`, which is still NULL.** It is allocated by
   `pfs_vncache_load()` from pseudofs's `MOD_LOAD` —
   `DECLARE_MODULE(pseudofs, ..., SI_SUB_EXEC, SI_ORDER_FIRST)` (`pseudofs.c:553`,
   SI_SUB_EXEC = 0x7400000). But every statically compiled pseudofs filesystem runs
   `pfs_init()` from `VFS_SET`'s `DECLARE_MODULE(..., SI_SUB_VFS, SI_ORDER_MIDDLE)`
   (`sys/mount.h:973-986`, SI_SUB_VFS = 0x4000000) — **three subsystems earlier**.
   `vfs_register()` calls `vfs_init` eagerly at registration (`vfs_init.c:517-519`).
   `MODULE_DEPEND` does not reorder SYSINITs for compiled-in code.
   With `pfs_vncache_hash == 0` (BSS) the loop runs exactly once and reads
   `*(void **)NULL` — `far = 0` exactly.

### Faulting instruction

Resolved against the CI `kernel.full` for `0f9119d6`. The running kernel is not
byte-identical (local symbol delta 0x60, proven twice independently: `lr − 0x60` lands on
the return address of the unique `bl pfs_purge` in `pfs_destroy`; `elr − 0x60` lands on the
only NULL-capable load in `pfs_purge`):

```
0xffff000000440120:  f8787909  ldr  x9, [x8, x24, lsl #3]
```

That is `SLIST_FIRST(&pfs_vncache_hashtbl[i])` at `pseudofs_vncache.c:311`, with
`x8 = pfs_vncache_hashtbl = NULL` and `x24 = i = 0`.

## Why CI and the VM did not catch this

An arm64 VM under UTM/qemu `-M virt` boots this exact kernel fine: a single generic PCIe
host means a single domain, no name collision, `linsysfs_init()` succeeds. The NULL hashtbl
is equally present there — only the trigger is hardware-specific. CI's arm64 boot job is
also `-M virt` (qemu has no BCM2712 model), so this class of failure is structurally
invisible to it.

## This is an inherited upstream bug

- **FreeBSD PR 278234** (2024-04-07, Graviton, domains 3 and 11, still New) is this exact
  bug in its pre-`D52038` form. kib acknowledged it and attached a draft patch
  (attachment 249811) that is essentially the per-domain layout below — never landed.
- `D52038` / `D52157` correctly downgraded the old `panic: pfs_add_node()` into a graceful
  `EEXIST`, but the vncache init-order hole silently converts that right back into a boot
  panic on the same hardware class.
- Any statically compiled pseudofs filesystem whose init fails before SI_SUB_EXEC panics
  identically. linsysfs is just the only one generating collision-prone dynamic names.

## Separately: upstream shipped a rebase artifact in 15.0

`linsysfs.c:297-302` contains a stray duplicated call:

```c
error = pfs_create_dir(dir, &dir, device, NULL, NULL, NULL, 0);
if (error != 0)
        goto out;
pfs_create_dir(dir, &dir, device, NULL, NULL, NULL, 0);   /* stray */
```

`cc70c7989bfb` added *one* error-checked call under the old pointer-returning API; the
coccinelle conversion `a2f08d0ddc29` converted it **and inserted a second copy**. Effect on
every platform including amd64: every PCI device directory is created twice, nested
(`/sys/devices/pci0000:00/0000:00:00.0/0000:00:00.0/vendor`), children of bridges nest under
the doubled dir, and the `scsi_host` `device` links — built from `new_path`, which appends
the name only once — dangle. Our amd64 kernels are shipping this today.

## Plan

**Land now (unbreaks boot, each piece independently correct):**

1. `pseudofs`: move the module to `SI_SUB_VFS, SI_ORDER_FIRST`. Verified safe —
   `hashinit(maxproc/4)` needs SI_SUB_KMEM (0x1800000) and `maxproc`, set by
   `init_param2()` from `initarm()` before `mi_startup()` runs any SYSINIT;
   `EVENTHANDLER_REGISTER` needs SI_SUB_EVENTHANDLER (0x1C00000). Shutdown ordering
   improves (pseudofs unloads after its consumers rather than before).
2. `pfs_purge()`: return early if `pfs_vncache_hashtbl == NULL`, **before** `mtx_lock` —
   the mutex is equally uninitialised. Semantically exact: no vnodes can exist before the
   cache does.
3. Remove the stray `pfs_create_dir()` at `linsysfs.c:301`.

**Fast-follow:** use `dinfo->cfg.domain` in the node name instead of the `"0000"` literal,
drop the now-dead `prefix` parameter, and fix `PCI_SLOT_NAME=%04d` → `%04x` in
`linsysfs_fill_uevent_pci` (wrong for domains >= 10). On single-domain amd64 `%04x` of
domain 0 renders byte-identical names — **zero amd64 behaviour change** — while the Pi's two
root ports become `0000:00:00.0` and `0001:00:00.0`.

**Known residual after step 1-3, before the fast-follow:** the second domain's subtree is
*omitted* rather than crashing — `goto out` skips that device and its children and the
recursion prints `linsysfs_run_bus: ... omitted from sysfs tree, error 17`. That is
upstream's designed degradation and is console-visible, not silent.

**Upstream:** submit 1-3 to FreeBSD (MFC/errata material for releng/15.0, since
`a2f08d0ddc29` shipped in 15.0); offer the per-domain layout as a rebase of kib's
attachment 249811 on PR 278234 rather than inventing a new one.

## Rejected

- **Drop LINSYSFS from arm64.** Nothing depends on it (`MODULE_DEPEND` runs
  linsysfs → linux_common, not the reverse), so it is a safe config change — but NextBSD
  has no kldload and no `.ko` tree, so it means no `/sys` *ever*, breaking libdrm/GPU
  enumeration. And it leaves the panic armed for whoever re-enables it. Not needed given
  how small the real fix is.
- **Only harden linsysfs's unchecked `pfs_create_*` return values.** Would not have
  prevented this panic: the fault happens inside `pfs_create_dir`'s own failure path before
  control returns to linsysfs, and the one site that can actually collide already checks its
  return. `M_WAITOK` makes ENOMEM impossible at the other sites.

## Verification

1. Boot the patched `kernel8.img` on the Pi 500+ — reaches login.
2. `pciconf -l` confirms two domains; `ls /sys/devices/` shows the layout.
3. amd64 boot unchanged; `/sys/devices/pci0000:00/0000:00:00.0/vendor` exists at one level
   (no nested duplicate).

pkgdemon:
Fixes #209.

The Pi 500+ stopped booting on current `main`. #208 compiled `LINSYSFS` into arm64 for the first time and `config/NEXTBSD-RPI5` does `include NEXTBSD`, so the Pi inherited it — the first time `linsysfs_init()` has ever run on arm64 hardware.

```
pfs_add_node: homonymous siblings: 'pci0000:00/0000:00:00.0' type 2
Fatal data abort:
  far: 0x0000000000000000
  esr: 0x0000000096000004
panic: vm_fault failed: 0xffff000000440180 error 1
Uptime: 1s
```

`esr 0x96000004` = level-0 translation fault on a **read**, `far` zero exactly. `elr`/`lr` resolve to `SLIST_FIRST(&pfs_vncache_hashtbl[i])` in `pfs_purge()` called from `pfs_destroy()` — pinned two ways against the CI `kernel.full`, both agreeing on a local symbol delta of `0x60`.

### Root cause

linsysfs names PCI nodes from a hardcoded `"0000"` domain prefix (`linsysfs.c:542`, `:290-294`), so the board's two PCIe host bridges — NVMe on pcie1, RP1 on pcie2, per patch 0013 — both render their root port as `0000:00:00.0`. `pfs_add_node()` correctly returns `EEXIST`; `pfs_create_dir()`'s error path then calls `pfs_destroy()` → `pfs_purge()`, which walks `pfs_vncache_hashtbl[]`. That table is allocated at `SI_SUB_EXEC` (0x7400000), but every statically compiled pseudofs filesystem registers from `VFS_SET()` at `SI_SUB_VFS` (0x4000000) and `vfs_register()` calls `pfs_init()` eagerly. `MODULE_DEPEND()` orders kldload, not SYSINITs.

### Patches

**0056 — pseudofs: create the vnode cache before the filesystems that use it.** Moves `DECLARE_MODULE` to `SI_SUB_VFS, SI_ORDER_FIRST` and adds a NULL guard in `pfs_purge()` *before* the `mtx_lock` (the mutex is uninitialised on the same paths). Preconditions verified: `hashinit()` needs SI_SUB_KMEM and `maxproc`, both set by `init_param2()` before `mi_startup()` runs any SYSINIT; `EVENTHANDLER_REGISTER()` needs SI_SUB_EVENTHANDLER. Shutdown ordering improves as a side effect.

**0057 — linsysfs: drop the duplicated `pfs_create_dir`.** Unrelated upstream rebase artifact found while reading the same function: `cc70c7989bfb` added one error-checked call, and the coccinelle conversion in `a2f08d0ddc29` converted it *and emitted a second copy*. Every PCI device directory is created twice and nested (`.../0000:00:00.0/0000:00:00.0/vendor`) and the `scsi_host` links dangle. **amd64 has been shipping this since 15.0.**

Also drops `boot.log`, a local qemu dry-run committed by accident in #208.

### Deliberately not fixed here

The second domain's subtree is still *omitted* rather than crashing, with `linsysfs_run_bus: ... omitted from sysfs tree, error 17` on the console. That is upstream's designed degradation and it is visible, not silent. Making linsysfs domain-aware is the fast-follow tracked in #209.

### Upstream

This is the boot-panic form of FreeBSD **PR 278234** (open since 2024, Graviton, domains 3 and 11). Both patches are written to go upstream verbatim and are MFC/errata material for releng/15.0.

### Testing

- Both patches `git apply --check` clean against pristine releng/15.0 and produce the intended result.
- **Not yet booted on hardware.** CI cannot cover this: qemu has no BCM2712 model, and the arm64 boot job runs `-M virt`, which has a single PCI domain and so cannot reproduce a multi-domain collision. Needs a `kernel8.img` swap on the Pi before merge.

https://claude.ai/code/session_014m6xNAWMJF6p7xPeGJKsJQ

### DIFF
--- a/boot.log
+++ b/boot.log
@@ -1,2 +1,0 @@
-spawn qemu-system-aarch64 -machine virt -accel tcg -cpu max -m 4096 -smp 2 -drive if=pflash,format=raw,unit=0,readonly=on,file=/usr/share/AAVMF/AAVMF_CODE.fd -drive if=pflash,format=raw,unit=1,file=vars.fd -drive if=none,id=hd0,format=raw,file=disk.img -device virtio-blk-device,drive=hd0 -netdev user,id=net0 -device virtio-net-device,netdev=net0 -display none -serial stdio
-qemu-system-aarch64: -drive if=pflash,format=raw,unit=0,readonly=on,file=/usr/share/AAVMF/AAVMF_CODE.fd: Could not open '/usr/share/AAVMF/AAVMF_CODE.fd': No such file or directory
--- a/patches/0056-pseudofs-create-the-vnode-cache-before-the-filesystems.patch
+++ b/patches/0056-pseudofs-create-the-vnode-cache-before-the-filesystems.patch
@@ -1,0 +1,98 @@
+From 3da5cf1109e66ca86e5296bbbc084aaf0eebb44e Mon Sep 17 00:00:00 2001
+From: pkgdemon <jpm820@gmail.com>
+Date: Thu, 10 Sep 2026 22:59:11 -0500
+Subject: [PATCH 1/2] pseudofs: create the vnode cache before the filesystems
+ that use it (#209)
+
+pfs_purge() walks pfs_vncache_hashtbl[] and locks pfs_vncache_mutex. Both are
+set up by pfs_vncache_load(), called from pseudofs' MOD_LOAD handler, which
+DECLARE_MODULE places at SI_SUB_EXEC (0x7400000). Every statically compiled
+pseudofs filesystem registers from VFS_SET() at SI_SUB_VFS (0x4000000) --
+three subsystems earlier -- and vfs_register() calls vfs_init(), and so
+pfs_init(), eagerly at registration time. MODULE_DEPEND() orders kldload, not
+SYSINITs, so it does not cover the compiled-in case.
+
+Any pfs_create_*() failure during a filesystem's pi_init() therefore reaches
+pfs_destroy() -> pfs_purge() with the hash table still NULL and the mutex
+still zeroed. pfs_vncache_hash is 0 at that point too, so the loop runs
+exactly once and reads *(void **)NULL.
+
+Measured on a Raspberry Pi 500+ (BCM2712) with LINSYSFS compiled in. The board
+has two PCIe host bridges, so linsysfs -- which names every node from a
+hardcoded "0000" domain prefix -- renders both root ports as 0000:00:00.0 and
+pfs_add_node() rejects the second:
+
+  pfs_add_node: homonymous siblings: 'pci0000:00/0000:00:00.0' type 2
+  Fatal data abort:
+    far: 0x0000000000000000
+    esr: 0x0000000096000004
+     lr: 0xffff00000043ecec
+    elr: 0xffff000000440180
+  panic: vm_fault failed: 0xffff000000440180 error 1
+  Uptime: 1s
+
+esr 0x96000004 is a level-0 translation fault on a read (WnR=0), far is zero
+exactly, and elr/lr resolve to the SLIST_FIRST(&pfs_vncache_hashtbl[i]) load
+in pfs_purge() called from pfs_destroy(). This is the boot-panic form of
+FreeBSD PR 278234, which D52038/D52157 turned from a pfs_add_node() panic
+into a graceful EEXIST -- graceful only if the teardown path survives.
+
+Moving the module to SI_SUB_VFS/SI_ORDER_FIRST is the fix: SI_ORDER_FIRST
+sorts before VFS_SET's SI_ORDER_MIDDLE at the same subsystem. Everything
+pfs_vncache_load() touches is ready by then -- hashinit() needs SI_SUB_KMEM
+(0x1800000) and maxproc, which init_param2() sets from initarm()/init386()
+before mi_startup() runs any SYSINIT, and EVENTHANDLER_REGISTER() needs
+SI_SUB_EVENTHANDLER (0x1C00000). Shutdown ordering improves as a side effect:
+module_shutdown runs in reverse registration order, so pfs_vncache_unload()
+now happens after its consumers rather than before them. kldload is unchanged.
+
+The NULL check in pfs_purge() is not redundant with the reordering; it states
+the invariant locally and keeps the teardown path safe for any future caller
+that runs earlier still. It has to precede the mtx_lock(), because the mutex
+is uninitialised on exactly the same paths as the table.
+
+Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>
+Claude-Session: https://claude.ai/code/session_014m6xNAWMJF6p7xPeGJKsJQ
+---
+ sys/fs/pseudofs/pseudofs.c         | 10 +++++++++-
+ sys/fs/pseudofs/pseudofs_vncache.c |  8 ++++++++
+ 2 files changed, 17 insertions(+), 1 deletion(-)
+
+diff --git a/sys/fs/pseudofs/pseudofs.c b/sys/fs/pseudofs/pseudofs.c
+index 7a4e674..4151f73 100644
+--- a/sys/fs/pseudofs/pseudofs.c
++++ b/sys/fs/pseudofs/pseudofs.c
+@@ -550,5 +550,13 @@ static moduledata_t pseudofs_data = {
+ 	pfs_modevent,
+ 	NULL
+ };
+-DECLARE_MODULE(pseudofs, pseudofs_data, SI_SUB_EXEC, SI_ORDER_FIRST);
++/*
++ * The vnode cache must exist before any statically compiled pseudofs
++ * filesystem is registered. VFS_SET() declares those at SI_SUB_VFS /
++ * SI_ORDER_MIDDLE and vfs_register() calls pfs_init() eagerly, so a
++ * pfs_create_*() failure there reaches pfs_destroy() -> pfs_purge()
++ * while pfs_vncache_hashtbl is still NULL. MODULE_DEPEND() orders
++ * kldload, not SYSINITs, so it does not help the compiled-in case.
++ */
++DECLARE_MODULE(pseudofs, pseudofs_data, SI_SUB_VFS, SI_ORDER_FIRST);
+ MODULE_VERSION(pseudofs, 1);
+diff --git a/sys/fs/pseudofs/pseudofs_vncache.c b/sys/fs/pseudofs/pseudofs_vncache.c
+index e58aced..34bccc3 100644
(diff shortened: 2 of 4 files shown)


pkgdemon:
## Verified on hardware — Pi 500+ boots

Booted this PR's `rpi5-kernel8-arm64` image on the Raspberry Pi 500+ from USB (`da0s1`), NVMe left
untouched as the fallback. Kernel `20260911-040906`, root on `/dev/ufs/ROOTFS`.

```
pfs_add_node: homonymous siblings: 'pci0000:00/0000:00:00.0' type 2
linsysfs_run_bus: pcib3 omitted from sysfs tree, error 17
linsysfs registered
```

That is precisely the intended outcome: the name collision still occurs, the losing device degrades
out with `EEXIST` (17), and **`linsysfs registered`** — init completes instead of panicking. Same
board, same kernel lineage, previously `panic: vm_fault failed` at `Uptime: 1s`.

## Hardware answers the open question in #209, with a correction

`pciconf -l`:

```
pcib1@pci0:0:0:0:  class=0x060400 vendor=0x14e4 device=0x2712
nvme0@pci0:1:0:0:  class=0x010802 vendor=0x1d97 device=0x5216
pcib3@pci2:0:0:0:  class=0x060400 vendor=0x14e4 device=0x2712
rp10@pci2:1:0:0:   class=0x020000 vendor=0x1de4 device=0x0001
```

The two PCI domains are **0 and 2**, not 0 and 1 as inferred in #209 from
`device_get_unit()`-as-domain. The host bridges are `pcib0` -> `pci0` and `pcib2` -> `pci2`; the
colliding pair is the root port on each, `pcib1@pci0:0:0:0` and `pcib3@pci2:0:0:0`. Both render as
`0000:00:00.0` once the domain is discarded, which is the collision.

Also now known: **the omitted subtree is the RP1 side** (`rp10@pci2:1:0:0` — Ethernet, USB, GPIO),
not the NVMe side. So until the domain-aware fast-follow lands, `/sys` is missing RP1 entirely.
Irrelevant to booting; relevant to anything doing sysfs device enumeration. Worth weighting the
fast-follow accordingly.

